PR #1, opened 2025-03-18
fix: update pyproject.toml to resolve packaging error

scott------:
The installation was failing due to setuptools discovering multiple top-level modules in the flat project layout. This commit updates the pyproject.toml by explicitly specifying:
  
• The standalone modules ("config.py", "server.py", "unity_connection.py") via the py-modules field.   • The "tools" directory as a package via the packages field.
  
These changes resolve the "multiple top-level modules discovered" error encountered during `uv pip install -e .` and ensure a clean, predictable package build.

### DIFF
--- a/Python/pyproject.toml
+++ b/Python/pyproject.toml
@@ -1,10 +1,21 @@
 [project]
-name = "unity"
+name = "unity-mcp"
 version = "0.1.0"
-description = "Add your description here"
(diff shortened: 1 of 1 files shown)


justinpbarnett:
Thanks!! Totally missed setting this file up 😅 it's been a busy day

PR #8, opened 2025-03-19
Update MCPEditorWindow.cs

Scriptwonder:
Appending to the config instead of overwrite


justinpbarnett:
Thanks!! This clears an issue off the board 🙏

PR #24, opened 2025-03-21
Feature/increase commands

justinpbarnett:
thanks @kuvkir !! excellent work. I added in a few more commands to be available ;)


kuvkir:
Nice, would probably make more sense to put execute_context_menu_item into the same category.

PR #26, opened 2025-03-21
added urp dependency

justinpbarnett:


glen-martin-tvmsi:
Hi, This breaks support for any Unity version below 6 because URP 17.0x is only support on Unity 6+ 
Added a commment here. Trying to use cursor as well
https://github.com/justinpbarnett/unity-mcp/issues/20#issuecomment-2746135581

PR #49, opened 2025-03-26
add text encoding process(base64) to 'View Script' Command for fixing timeout issue.

flashwade03:
Hi great developer @justinpbarnett !

First of all, I would like to thank you for making a great start to a great project. 

I was debugging a timeout issue with the view script command that some people are currently experiencing, and I would like to point out that the cause was the long script and the presence of non-English comment characters in the script. So, the method I chose was to encode it in base64 type and send it back and forth between server<->client and decode it to use, and I confirmed that the problem was solved to a large extent. I would like to apply this problem to the update script as well, but I have not analyzed the entire code yet, so I will share it here for now.

If I find any other improvements or add my own good features in the future, I will share them with you.

Thank you again for your great work.

Have a nice day.

### DIFF
--- a/Editor/Commands/ScriptCommandHandler.cs
+++ b/Editor/Commands/ScriptCommandHandler.cs
@@ -1,10 +1,10 @@
-using UnityEngine;
-using UnityEditor;
 using System;
 using System.IO;
-using System.Text;
 using System.Linq;
+using System.Text;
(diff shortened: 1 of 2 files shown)


SoundGuy:
you probably should update the version to a higher value than 0.1.4 if you are merging PRs so that we can update the package in unity package manager

PR #52, opened 2025-03-27
Add UnityMCP.Editor.asmdef

Alex-Rachel:
Fixed that some versions of Unity cannot display the MenuItem of Window/Unity MCP


dsarno:
Thank you for your contribution, and sorry for the delay, we're just catching up now.  I believe elsewhere we've decided that these menu display issues are no longer a problem in the latest versions of the mcp.  Please let us know if that's not the case.  I'm going to close this for now.

PR #53, opened 2025-03-27
add Optional remaings

akjava:
Fix: Add missing Optional type hint to parameters


bhadrik:

dsarno:
Going to close for now as it's a bit stale and I'm not quite clear of the utility. Happy to reopen with more explanation.

PR #55, opened 2025-03-27
add using FindTypeInLoadedAssemblies

flashwade03:
HI.

I think it would be better to add 'FindTypeInLoadedAssemblies' method as a way to retrieve custom class types when using ModifyObject tool.


bhadrik:

Scriptwonder:
Going to close this PR because the current code structure seems to be very different from the one you edited before. Is there any specific reason or scenarios that this will be needed?

PR #66, opened 2025-03-31
Fix uv package manager installation command for Windows.

portlek:


README.md
```diff
@@ -41,7 +41,7 @@ uv is a Python package manager that simplifies dependency management. Install it
 - **Windows**:
 
   ```bash
-  powershell -c "irm https://astral.sh/uv/install.ps1 | iex"
+  powershell -ExecutionPolicy ByPass -c "irm https://astral.sh/uv/install.ps1 | iex"
```
jsjtxietian, line 44: To be honest after I found that the origin comand did not work, I choose to use pip, very convenient :)

portlek: this is just as simple as this actually:
`irm https://astral.sh/uv/install.ps1 | iex`
you don't really want* to run the whole thing. you just need to run the powershell administrator then run the only the script part. but people might get confused by it because it doesn't really say it should be run as administrator.

justinpbarnett:
Hey I appreciate you! Ended up overhauling the readme and simplified the uv installation steps to just `pip install uv` and a link to the docs for anything else in PR #83 

Really appreciate your willingness to jump in on this tho.

PR #80, opened 2025-04-08
Rename Python to Python~ to avoid unity importing them

jsjtxietian:
Fixes https://github.com/justinpbarnett/unity-mcp/issues/60

I choose to rename the folder to prevent unity from importing them.

Also I tried to relocate the pycache file to somewhere else, but there is also egg-info file that needs relocate, so I thought we can rename the folder to fix the issue once for all. Also we can delete all the meta files in this way


justinpbarnett:
This is a great direction. Was about to pull your fix in, but I ended up just totally overhauling the python server install process in PR #82, so now it installs separately as a totally different application. Gonna close this out, but really appreciate your willingness to jump in here! ❤️

PR #84, opened 2025-04-09
Feature/advanced script checking

Scriptwonder:
Use Roslyn to enable better script checking
Add one more Response type to alert LLM that current script generated has a list of errors
Use Preprocessing Directives to wrap the Roslyn function as optional choice


bhadrik:

Scriptwonder:
Will implement advanced script checking method instead of this PR.

PR #86, opened 2025-04-09
Remove python server status UI

jsjtxietian:
Fixes https://github.com/justinpbarnett/unity-mcp/issues/85

We can also add the status back, but:
1. we can't really do something about it
2. the status is refactored into `McpStatus` and it's a mix between python server's status along with something else, the code needs some more change to reflect the python server status

But I suppose add the status back is ok too


justinpbarnett:
Just pushed an update to change the status to be install status. Will need a bit more code to actually check for the python connected status and don't quite have the bandwidth at the moment. Appreciate your PR!

justinpbarnett:
Resolved (for now) in PR #87

PR #108, opened 2025-04-19
Add Dockerfile

michaelnugent:


Scriptwonder:
Hi! Thanks for the work! Can you check if this is still runnable in current version? I will close and merge it since it looks alright to me, but just wanna double check.

PR #131, opened 2025-05-23
Add support for creating physics material assets

Sunjnn:
Hi, I have added a new function `ApplyPhysicsMaterialProperties` to the project, which facilitates the creation of physics material assets.

### DIFF
--- a/UnityMcpBridge/Editor/Tools/ManageAsset.cs
+++ b/UnityMcpBridge/Editor/Tools/ManageAsset.cs
@@ -177,6 +177,14 @@ namespace UnityMcpBridge.Editor.Tools
                     AssetDatabase.CreateAsset(mat, fullPath);
                     newAsset = mat;
                 }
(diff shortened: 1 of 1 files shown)


Scriptwonder:
Have you run different use cases for this method (E.g. do you prompt specifying PhysicsMaterial and something similar)?  Looks cool!

Sunjnn:
I have tried running use cases where I specify PhysicsMaterial, and it works.

Scriptwonder:
I have tried it and it encounter several issues:

1. There is no PhysicsMaterial but PhysicMaterial in the older Unity version, I assume you are using Unity 6? I am running Unity 2022.3.60f1, and multiple errors showed up requiring me to clear that conflict. 
2. For the set up of the three different properties, you might need to update the python file to let the output also follow the Pascal Case. This is something that is bugging most of the tools, and I am still thinking of a way to solve this.

PR #139, opened 2025-06-07
Unity MCP  One-Click Setup for Public Distribution & Claude Code bridge

6years2Roi:
Description:
   - Complete self-contained distribution package
   - Universal launcher with combined server + bridge
   - Smart Git fallback system (works with or without Git)
   - Auto Python environment setup
   - WSL/Linux compatibility with network IP detection
   - Reduces setup time from 30 minutes to 30 seconds
   - Suitable for public project distribution


dsarno:
@6years2Roi Same here. These are just too large to handle. If you want to contribute, please pick a smaller targeted piece to improve and submit a lean, reviewable PR. Thanks again!

PR #144, opened 2025-06-22
Support discovering tools defined in unity project.

Reekin:
Add IUnityMcpTool interface and make the bridge discover the tools defined in unity project.
And to unify where tools are defined, I moved the pre-defined tools information from server to the package.
By the way, supported listening messages from WSL, which is useful for Claude Code.


dsarno:
@Reekin Thanks for this. Auto‑discovering MCP tools with an IUnityMcpTool interface—is solid and would make it easier for people to add tools.

But we run into issues because the repo has evolved a bunch in recent weeks and as a very small team we're only now reviewing older PRs. This PR is pretty large (~59 files, +3k/−700 lines), targets master instead of main, and mixes multiple changes (Unity registry, “list tools,” Examples, plus WSL/server files). It also touches hot spots like the bridge and server that have changed on main, so it would be a noisy merge. To keep the repo tidy, I’m going to close this for now.  ￼

If you’d like to revive it, a small, rebased PR focused on just:
	•	the IUnityMcpTool interface,
	•	a minimal dynamic discovery/registration in the Unity package,
	•	and a simple “list tools” command,

…would be easy to review. Please leave out the WSL/server tree, lockfiles, and the large Examples folder—those can live in docs or a separate sample repo. We’ll happily review a slimmer, rebased version against main.

PR #167, opened 2025-07-21
Feature/asmdef fixer tool

cjpxyz:
A tool to automatically analyze and fix .asmdef references for a given C# script.


.gitignore
Scriptwonder, line 1: Why are we adding this in the current project? Unity-mcp resides in the package and will not include these typical files/folders in either the server/client side.

UnityMcpBridge/Editor/UnityMcpBridge.cs
Scriptwonder, line 1: Where is FixAsmdefReferences.cs?

UnityMcpServer/src/tools/__init__.py
Scriptwonder, line 1: I think because you are using a different LF (Possibly LF and possibly written with a LLM), it confused git diff and generate a overhaul of add/delete. Try change that so it would be more helpful to view and merge.

dsarno:
Thanks for this. The idea—a tool to scan and fix .asmdef references—is useful for Unity projects that drift into broken assembly refs. I'm not 100% convinced it needs to be part of an mcp since it's so single purpose. But it's a good utility idea. Maybe for its own small repo?
 
Also this PR is behind our main branch, targets master, and mixes in unrelated changes (lockfile and a bigger .gitignore sweep). It also edits hot files (the Unity bridge) and a reviewer notes at least one missing file reference (FixAsmdefReferences.cs). Given the drift and conflicts, I’m going to close this for now.

PR #171, opened 2025-07-27
Feature/enhanced error handling

usexless:


dsarno:
Thanks for pushing on better timeouts and clearer errors — that’s useful. We actually addressed much of that in our recent updates so we agreed it was necessary.

This PR is very large (31 files, +9k/−254 lines), includes committed log files, and bundles a sweeping refactor (new exception/timeout/validation/logging layers, remote connection features, health endpoint, etc.). It also targets master while our default branch is main. To keep the repo lean and predictable, I’m going to close this for now. But let us know if you want to try another more targeted PR. If so please spend more time on the description, explanation and justification when you submit.  Thanks ￼

PR #172, opened 2025-07-27
Feature/custom tool creation gui

usexless:


Scriptwonder:
Hi there! Thanks for the massive effort put into this PR(and the previous one), I can see you put a lot of work into improving this work. However, I have found some serious issues that need fixing before we can merge this.
1. custom_tool_creator.py let users inject arbitrary unity commands without any validation, considering we will do remote hosting in the future, this is a major security flaw. east_setup is also one such thing that can lead to command injection by pass user input directly to subprocess commands. 
2. The AI coding pattern is somehow obvious, in both the readme and the py files. Different timeout systems doing similar things, multiple overlapping logging frameworks, and some complete rewrite of core systems, requiring old users to went through complex installation process again.
3. The core issue you mentioned: **infinite waiting for responses...** How can I reproduce this issue? This issue never happened during my usage, and I am genuinely curious on this issue especially when right now we are detailing our response message to include more details of failure on the Unity side so LLM can understand (Speaking of which, a lot of changes are on the python side in this PR, which I would like to know more on the reason for this change)
4. In the end, please split this into smaller PRs, as this is way too much to review safely. I would suggest you fix the security fixes, trying to simplify the code structure, remove duplicate functionality, and split them into different PRs that can piece together. 

Appreicate the effort! Would you be open to breaking this down and tackling the issues first? Happy to help review smaller chunks based on new features.

dsarno:
I also thank you for the effort but we can't accept these super huge AI-coded PRs, they're big and hard to review. If you want to contribute please pick something small to start, and spend some more time on the explanation and justification when you submit. Let us know if you want to try a small one.  We appreciate your interest!

PR #185, opened 2025-08-02
Fix on focus config check

msanatan:
Ref: https://discord.com/channels/1394050967917428756/1396128287771590757/1401040388541648998

Prevents the log spam, slightly more efficient check when the Window is back in focus.

This change was added very recently in #175, maybe check with the author first before merging this in, it continues to work for my connection but I'm not sure if this unwittingly breaks something. Feels safe so far


Scriptwonder:
@dsarno Looks good to me! Let me know what you think on this.

Issue #181, opened 2025-07-31
System.Exception: No available ports found in range 6400-6500 why?


RouderSky:
unity 2021.3.30f1

System.Exception: No available ports found in range 6400-6500
  at UnityMcpBridge.Editor.Helpers.PortManager.FindAvailablePort () [0x0008d] in J:\PythonProject\unity-mcp\unity-mcp\UnityMcpBridge\Editor\Helpers\PortManager.cs:85 
  at UnityMcpBridge.Editor.Helpers.PortManager.GetPortWithFallback () [0x00033] in J:\PythonProject\unity-mcp\unity-mcp\UnityMcpBridge\Editor\Helpers\PortManager.cs:43 
  at UnityMcpBridge.Editor.UnityMcpBridge.Start () [0x0003d] in J:\PythonProject\unity-mcp\unity-mcp\UnityMcpBridge\Editor\UnityMcpBridge.cs:109 
  at UnityMcpBridge.Editor.UnityMcpBridge..cctor () [0x0002b] in J:\PythonProject\unity-mcp\unity-mcp\UnityMcpBridge\Editor\UnityMcpBridge.cs:84 
UnityEditor.EditorAssemblies:ProcessInitializeOnLoadAttributes (System.Type[])

netstat -an | findstr 64  
  TCP    0.0.0.0:49664          0.0.0.0:0              LISTENING
  TCP    127.0.0.1:1408         127.0.0.1:6469         SYN_SENT
  TCP    127.0.0.1:8988         127.0.0.1:14864        ESTABLISHED
  TCP    127.0.0.1:13646        127.0.0.1:13647        ESTABLISHED
  TCP    127.0.0.1:13647        127.0.0.1:13646        ESTABLISHED
  TCP    127.0.0.1:13648        127.0.0.1:13649        ESTABLISHED
  TCP    127.0.0.1:13649        127.0.0.1:13648        ESTABLISHED
  TCP    127.0.0.1:14864        127.0.0.1:8988         ESTABLISHED
  TCP    127.0.0.1:56436        0.0.0.0:0              LISTENING
  TCP    127.0.0.1:56864        0.0.0.0:0              LISTENING
  TCP    192.168.3.2:1264       163.177.58.174:6506    ESTABLISHED
  TCP    [::]:49664             [::]:0                 LISTENING
  UDP    0.0.0.0:49664          *:*
  UDP    127.0.0.1:64729        *:*
[...]

Issue #190, opened 2025-08-08
python version must be 3.12.11


Awakeeeee:
tutorial said python require 3.12 or newer, but in my case 3.13 won't work ( Unity logs everything is OK but Cursor said no tools are loaded)
mcp server trys to download a specified python version (3.12.11) and lots of other requirements on start.
looks like server only works under this version of python.

dsarno:
@Awakeeeee I'd be curious if you got the same issue with the update I just proposed as a PR, which includes stuff to explicitly fix any python port and version issues. You can test it by changing one line in your manifest.json file in Unity. I put the instructions in [this message](https://discord.com/channels/1394050967917428756/1400341207837708360/1403507909282758696) in the repo's Discord

dsarno:
This should be fixed in the unity-mcp version `2.0.0` (via PR https://github.com/CoplayDev/unity-mcp/pull/192 ). Let us know if you have any issues, and thank you.

Issue #103, opened 2025-04-16
Port in use


wacky444:
After using mcp for a while in windows since it starts the server automatically each time unity recompiles the code it ends up throwing port already in use and the mcp server doesn't respond.

Maybe an option to disable auto start.

I'am using copilot in visual studio code.

Prikalel:
Yes i have the same trouble
Using cline in vs code

SujanMishra:
yes same thing 
Port 6400 is already in use. Ensure no other instances are running or change the port.
UnityEngine.Debug:LogError (object)
UnityMcpBridge.Editor.UnityMcpBridge:Start () (at Library/PackageCache/com.justinpbarnett.unity-mcp/Editor/UnityMcpBridge.cs:91)
UnityMcpBridge.Editor.UnityMcpBridge:.cctor () (at Library/PackageCache/com.justinpbarnett.unity-mcp/Editor/UnityMcpBridge.cs:53)
UnityEditor.EditorAssemblies:ProcessInitializeOnLoadAttributes (System.Type[])

RandomNameQ:
same

zhengjianhua1:
same, close vscode then start bridge is ok

when vscode is opening

The close method of TCPListener in Unity code does not work

when close Unity

netstat -ano | find "6400"
  TCP    127.0.0.1:6400         0.0.0.0:0              LISTENING       39392

 Unity's pid is  39392, but the process does not actually exist

radaelilucca:
Same issue. Looks like vscode is starting its own server on same port.

RainlandLuo:
same issue happens after worked a while, the error shows in unity console

Efe-Oral:
Did you find a solution for Error:Port 6400 is already in use. Ensure no other instances are running or change the port.

manntera:
Hi, I'm a Japanese user and I'm using AI to help me write this in English, so I apologize in advance if anything sounds strange or incorrect.

I encountered the same issue described in [Issue #103](https://github.com/justinpbarnett/unity-mcp/issues/103) regarding UnityMCP always using port 6400, which was already in use on my machine.

When I checked using the following commands:

```
netstat -an | findstr 6400
netstat -aon | findstr :6400
tasklist | findstr 3988
```

I found that port 6400 was being used by `svchost.exe` (PID 3988), so UnityMCP could not start correctly due to a port conflict.

[...]

Issue #92, opened 2025-04-10
Issue with Cursor


imumoccupancium:
Alright, so I tried installing MCP git from the repository into my Unity Packet Manager, everything is fine and perfect in the Unity side. However, when I opened the Cursor, it's showing this "client closed". Though I did everything what the instructions says. 

![Image](https://github.com/user-attachments/assets/9a383e59-40f6-49b7-b0c3-1f568e5647df)

![Image](https://github.com/user-attachments/assets/466b6874-7839-478d-8e80-f5de61cf1286)

justinpbarnett:
Ah this is probably due to your folder "Richard Andrew Recio" having spaces in it.

The quick fix is to add some escaped double quotes around the directory section of your config. So your path would end up looking like this in the config:
```json
"\"C:\Users\Richard Andrew Recio\AppData\Local\Programs\UnityMCP\UnityMcpServer\src\""
```

justinpbarnett:
Duplicate of #21

imumoccupancium:
It hasn't fixed the problem, the problem remains the same after debugging with your solution.

K53N0:
> It hasn't fixed the problem, the problem remains the same after debugging with your solution.

Check Cursor logs in the bottom panel, output tab.
There are probably some clues there. This happened to me and the problem was that the path for uv wasn't correct, so I had to manually find and fix it in the mcp json file: 

// ... existing code ...
    },
    "unityMCP": {
      "command": "C:\\Users\\kseno\\AppData\\Roaming\\Python\\Python313\\Scripts\\uv.exe",
      "args": [
        "--directory",
        "C:\\Users\\kseno\\AppData\\Local\\Programs\\UnityMCP\\UnityMcpServer\\src",
        "run",
        "server.py"
      ]
    }
  }
}

This obviously assumes you installed uv or whatetever other necessary import via pip install

sneezeparty:
With the latest version (2.0), I fixed this error by turning off the UnityMCP Bridge.  The only reason I turned it off was to see what effect it would have on console output regarding the error...  but when I hit the reload button in Cursor to try reconnecting to the MCP server, the light turned green.  And now it's working.

dsarno:
Hi all, try again using the just-released `2.0.0` version of the MCP, which has major improvements to setup and mcp bridge/server configuration. You should reinstall the package in Unity because there's a new package name and address.  I'm going to close this but can reopen if you still see problems.

```
1. Open your Unity project.
[...]

PR #192, opened 2025-08-08
Harden MCP Bridge: reliable reloads, sticky port, embedded server, cleaner Auto‑Setup UI

dsarno:
## Overview
This PR hardens the Unity MCP Bridge and streamlines the end‑to‑end setup. It removes the old Git‑based server installer in favor of an embedded Python server, reduces perceived downtime across domain reloads, and improves the Unity Editor UX for configuration and troubleshooting.

- Related: Addresses https://github.com/CoplayDev/unity-mcp/issues/181 and similar

## Behavior Changes (Before → After)
- Ports could hop on reload (e.g., 6401/6402) → Reuses the same per‑project port (e.g., 6400) with micro bind‑retries
- Bridge sometimes didn’t restart after compiles → Reliable auto‑restart with short retry loop and socket options for all tools. 
- Editor window could show stale status → UI reflects live bridge state; status survives domain reloads
- Client setup noisy/manual → Cleaner UI with one “Auto‑Setup” button, clear status, and fewer logs by default
- Server installed via Git after package import → Python server is embedded in the package; no network fetch for users

## Why
- Eliminate port hopping and connection gaps during domain reloads
- Make installation self‑contained and robust on first import and upgrades
- Provide a single, obvious way to repair local Python environments (.venv mismatch, system Python version changes)
- Warn if python not detected, offer install link
- Increase clarity in the Editor window while gating verbose logs behind a global toggle

### Bridge lifecycle: `UnityMcpBridge/Editor/UnityMcpBridge.cs`
- Static init starts bridge quickly; writes a heartbeat file indicating `unity_port` and `reloading` state
- Start/Stop guarded by locks; short micro‑retry loop on `AddressAlreadyInUse`
- Socket options for resiliency: `ReuseAddress` and `Linger(0)` to reduce TIME_WAIT binds
- Periodic heartbeat written at ~0.5s to `~/.unity-mcp/unity-mcp-status-<hash>.json`
- Listener errors handled gracefully; ping fast‑path for health checks

### Port management: `UnityMcpBridge/Editor/Helpers/PortManager.cs`
- Prefers persisted project port (case‑insensitive project path match)
- Returns stored project port even if transiently busy; Start uses micro‑retry to avoid port flips
- Reads/writes both hashed per‑project file and legacy file for back‑compat

### Embedded server installer: `UnityMcpBridge/Editor/Helpers/ServerInstaller.cs`
- Install path is user‑writable (macOS: `~/Library/Application Support/UnityMCP`), no Git required
- New `RepairPythonEnvironment()` deletes `.venv` and runs `uv sync` to rebuild a clean env
- Robust UV detection (common paths, PATH scan, `which uv`), plus `EditorPrefs` override (`UnityMCP.UvPath`)
- Fallback logic: if installed path missing, uses embedded/dev source to repair

### Editor UI: `UnityMcpBridge/Editor/Windows/UnityMcpEditorWindow.cs`
- 2×2 layout with consistent sizing; header‑level “Show Debug Logs” toggle (globally gates verbose logs)
- “Server Status” shows Installed/Installed (Embedded)/Not Installed
- One‑click “Repair Python Env” with tooltip; fewer logs by default
- Python detection warning with a link to official installers: https://www.python.org/downloads/
- Dev‑mode aware path resolution via `FindPackagePythonDirectory()` so dev clones don’t appear misconfigured (in other words, all you have to do to swap to a dev clone is change the `manifest.json` file in Unity)

### Python server (vendored)
- Embedded at `UnityMcpBridge/UnityMcpServer/src`
- Remove `.python-version` pin; set `requires-python = ">=3.10"` in `pyproject.toml`
- `uv.lock` updated

## Stability/UX Improvements
- Sticky per‑project port across reloads and play‑mode transitions
- Bridge restart is fast and reliable (micro‑retry loop); reduced perceived downtime for clients
- Status survives domain reloads via heartbeat and UI refresh logic
- Clear feedback for missing Python with direct link to installer
- “Repair Python Env” fixes stale/mismatched venvs without manual steps

## Compatibility
- Backwards‑compatible: legacy port files still supported
- No public tool API/schema changes
- UPM distribution includes embedded server; local dev continues to work with file: manifests

## Risks & Mitigations
- UV discovery may miss unusual installations → PATH and `which` scan + `EditorPrefs` override
- Rare bind failures after reload → micro‑retries in Start; prefers stored port to avoid flip‑flop
- Heartbeat write is best‑effort; failures are non‑fatal

## How to Verify (Quick)
1. Import the package (UPM) and open the MCP window; confirm bridge starts and shows the port (e.g., 6400)
2. Make a trivial script change to trigger a domain reload → bridge restarts; port unchanged
3. Click “Repair Python Env” → completes without error; UV output errors appear in Console if any
4. From a client, run `read_console` across reloads → remains responsive

## Follow‑ups (Optional)
- Consider a background auto‑repair if `.venv` import errors are detected at runtime
- Optional: expose a manual UV path field next to the Repair button if override is frequently needed

## Links
- PR comparison: https://github.com/CoplayDev/unity-mcp/compare/main...dsarno:unity-mcp:feat/bridge-stability?expand=1
- Python installers: https://www.python.org/downloads/


dsarno:
**From GPT 5 Pro:** Here’s a focused review of **PR #192** (“Unity MCP: bridge stability + embedded server”), based on the PR thread and the key commits/diffs. I’ve grouped notes into **what looks great**, **risks/edge cases**, and **actionable suggestions**, then closed with a quick **test plan** you can run before merge.

Note from PR author @dsarno : Great feedback, not sure it's all necessary for a merge, but I can work on some of it for a follow-up PR unless anyone feels strongly. _Edit: Ended up hardening the reload and retry/debounce functionality, much better than in the original commit, see comment below._

---

## What’s solid

1. **Sticky, per‑project ports + brief release wait**
   `PortManager` now prefers the previously stored port, waits briefly during domain reloads, and only falls back when necessary. This directly targets the “port flipping on recompiles” pain. Nice touch adding a short wait and a quick socket probe to reduce churn. ([[GitHub](https://github.com/CoplayDev/unity-mcp/commit/32274a396586988addd6f7babc89509b826f17f6)][1])

2. **Embedded Python server instead of Git-based installer**
   Switching `ServerInstaller` to copy from an embedded `UnityMcpServer/src` removes network flakiness and makes installs deterministic. The code paths to detect embedded/dev layouts plus status/UI tweaks are a clear simplification. ([[GitHub](https://github.com/CoplayDev/unity-mcp/commit/a65f10383ab5ec381adcbb6f8167f9e5c9d3a385)][2])

3. **Repair flow for Python env + better uv discovery**
   Adding `RepairPythonEnvironment()` (delete `.venv`, unpin `.python-version`, run `uv sync`) and robust `uv` path discovery (EditorPrefs override + several common locations + PATH probing) will save users time when Python breaks. Also moving macOS install location to Application Support is appropriate for a user-writable target. ([[GitHub](https://github.com/CoplayDev/unity-mcp/commit/f24e124c1565036ec50d80007a43aa7e2daf112e)][3])

4. **Small, user-facing polish**
   Button rename (“Bind to Clients”) and other editor status/UX cleanups make the window less confusing. ([[GitHub](https://github.com/CoplayDev/unity-mcp/commit/2f387d3417a7b85c901fa0f19e2aee64d35e35c3)][4])

5. **Scope and intent match the PR description**
   The PR’s emphasis on recompile/play-mode resilience, start/stop locking, sticky ports, and vendoring the server is consistent with the diffs/commits reviewed. ([[GitHub](https://github.com/CoplayDev/unity-mcp/pull/192)][5])

---

## Risks & edge cases to consider

1. **Main-thread stalls from networking checks**
   `IsPortUsedByUnityMcp` does `TcpClient.ConnectAsync(...).Wait(100)`. Even at 100 ms, this is a synchronous wait that can hitch the editor if called on the main thread during domain reload or window repaint. Consider a non-blocking pattern or doing this in a background task with a short timeout. ([[GitHub](https://github.com/CoplayDev/unity-mcp/commit/32274a396586988addd6f7babc89509b826f17f6)][1])

2. **Heuristic for “port is MCP” may produce false positives**
   Right now the probe only checks that *something* connects on the port. If another local process temporarily owned the same port, you might incorrectly “wait for MCP” and stall. A minimal handshake (see suggestion #2) would make the check definitive. ([[GitHub](https://github.com/CoplayDev/unity-mcp/commit/32274a396586988addd6f7babc89509b826f17f6)][1])

3. **Sticky-port logic during hot transitions**
   You’ve reinforced “prefer stored port” (micro-retry then bind) to defend against swap-flapping—which is great—but keep an eye on scenarios where another process snipes the port between stop/start. Ensure the bind path has a bounded retry/backoff and logs *why* it finally moved. ([[GitHub](https://github.com/CoplayDev/unity-mcp/commit/f4bc7cd4fded943881c9e7d16271a802eb3ae400)][6])

4. **Cross‑platform install paths / permissions**
   macOS path change to `Application Support` is good. Verify Windows paths (e.g., `%LOCALAPPDATA%\\Programs\\UnityMCP\\UnityMcpServer\\src`) are consistently used across UI messaging and any auto-config writers. Mixed, dangling legacy paths can confuse users after upgrading from the Git-based installer. ([[GitHub](https://github.com/CoplayDev/unity-mcp/commit/f24e124c1565036ec50d80007a43aa7e2daf112e)][3])

5. **Package bloat & import times**
   Vendoring the Python server into the Unity package increases content under UPM. Keep an eye on import times and .meta noise. If this grows further, consider excluding non-essential dev artifacts from UPM distribution (e.g., large caches, test data). ([[GitHub](https://github.com/CoplayDev/unity-mcp/commit/a65f10383ab5ec381adcbb6f8167f9e5c9d3a385)][2])

---

## Actionable suggestions (fast wins)

1. **Make the port probe non-blocking**
   Replace `ConnectAsync(...).Wait(100)` with a true async/timeout approach off the main thread. Two options:

* Schedule the probe via `EditorApplication.delayCall` and continue the binding flow only after it returns.
* Use `Task.WhenAny(connectTask, Task.Delay(100))` and avoid blocking waits entirely.
  This removes potential 100–1500 ms UI hitches during reloads. ([[GitHub](https://github.com/CoplayDev/unity-mcp/commit/32274a396586988addd6f7babc89509b826f17f6)][1])

2. **Add a lightweight identification handshake**
   Have the server expose a trivial “hello” (e.g., send a single line and expect `UNITY_MCP\n`, or an HTTP `/healthz` if you’re using HTTP). Then `IsPortUsedByUnityMcp` can attempt that handshake and positively identify MCP vs random listeners. Document the handshake contract in the repo for future tooling. ([[GitHub](https://github.com/CoplayDev/unity-mcp/commit/32274a396586988addd6f7babc89509b826f17f6)][1])

3. **Constrain/telemetry retry behavior on bind**
   Where you “prefer stored port,” add:

* A small, capped retry loop with exponential backoff (e.g., 3 attempts at 50/100/200 ms).
* A clear final log that explains whether you stuck to the stored port or had to pick a new one.
  This will make “why did it change?” obvious in user reports. ([[GitHub](https://github.com/CoplayDev/unity-mcp/commit/f4bc7cd4fded943881c9e7d16271a802eb3ae400)][6])

4. **Unify and gate logs**
   You already have a “Show Debug Logs” toggle in the editor window. Route all bridge/server‑management logs through a single helper (prefix `[UNITY‑MCP]`, level‑aware). This keeps normal use quiet and gives power users precise breadcrumbs when debugging. ([[GitHub](https://github.com/CoplayDev/unity-mcp/commit/f24e124c1565036ec50d80007a43aa7e2daf112e)][3])

5. **Upgrade/migration note for legacy installs**
   On first run after update, detect the legacy Git-based install location. If found, show a one-time banner:

* “Embedded server is now used.”
* Offer buttons: “Open Folder,” “Remove Legacy Copy,” “Dismiss.”
  It reduces confusion and orphaned data. ([[GitHub](https://github.com/CoplayDev/unity-mcp/commit/a65f10383ab5ec381adcbb6f8167f9e5c9d3a385)][2])

6. **CI sanity checks (optional but useful)**

* A headless test that spins the Python server (via `uv`) and hits the handshake endpoint.
* A portability test to validate `FindUvPath()` candidates on macOS/Linux/Windows naming variants.
  These catch regressions before users do. ([[GitHub](https://github.com/CoplayDev/unity-mcp/commit/f24e124c1565036ec50d80007a43aa7e2daf112e)][3])

---

## Targeted code nits

* **Port registry naming:** You added hashing + legacy filename handling—good. Ensure path comparisons are case‑insensitive on Windows (looks like you covered this in follow-ups). Add a brief comment where you compute the hash stating the input (full project path, normalized). ([[GitHub](https://github.com/CoplayDev/unity-mcp/commit/32274a396586988addd6f7babc89509b826f17f6)][1])
* **Thread.Sleep in editor code:** `WaitForPortRelease` uses `Thread.Sleep(step)` in a loop. If there’s any path where this runs on the main thread, replace with async delay or editor updates to avoid UI stalls. ([[GitHub](https://github.com/CoplayDev/unity-mcp/commit/32274a396586988addd6f7babc89509b826f17f6)][1])
* **Timeout values:** Current 100 ms probe and \~1500 ms total wait are reasonable. Document the rationale and make them constants with names like `kProbeTimeoutMs` / `kPortReleaseWaitMs` to ease future tuning. ([[GitHub](https://github.com/CoplayDev/unity-mcp/commit/32274a396586988addd6f7babc89509b826f17f6)][1])

---

## Quick test plan (manual)

Run these in both **macOS** and **Windows**:

1. **Fresh install path**

   * Remove existing installs; import the package.
   * Open *Window → Unity MCP* and confirm **Installed (Embedded)** status and no network fetching. ([[GitHub](https://github.com/CoplayDev/unity-mcp/commit/a65f10383ab5ec381adcbb6f8167f9e5c9d3a385)][2])

2. **Recompile / domain reload stability**

   * Start MCP Server, connect a client.
   * Trigger a script recompile (edit C# file) and toggle Play Mode on/off several times.
   * Confirm: port remains the same; reconnects are automatic; no “port already in use” flapping. ([[GitHub](https://github.com/CoplayDev/unity-mcp/commit/f4bc7cd4fded943881c9e7d16271a802eb3ae400)][6])

3. **Port collision behavior**

   * Manually bind a dummy listener to the stored port.
   * Start the bridge and verify: brief wait occurs; if non‑MCP process owns it, you don’t stall indefinitely and logs explain the fallback. ([[GitHub](https://github.com/CoplayDev/unity-mcp/commit/32274a396586988addd6f7babc89509b826f17f6)][1])

4. **Repair flow**

   * Corrupt `.venv` or pin an incompatible `.python-version`.
   * Click **Repair**; confirm `uv sync` runs and the server starts cleanly. ([[GitHub](https://github.com/CoplayDev/unity-mcp/commit/f24e124c1565036ec50d80007a43aa7e2daf112e)][3])

5. **UI/wording**

   * Verify “Bind to Clients” flows still update configs as expected and status is accurate. ([[GitHub](https://github.com/CoplayDev/unity-mcp/commit/2f387d3417a7b85c901fa0f19e2aee64d35e35c3)][4])

---

## Overall take

The direction is right: stickiness across reloads, deterministic installs, and practical recovery tools. Before merge, I’d address the **main‑thread waits**, add a **positive MCP handshake**, and make the **bind‑retry behavior** explicit and well‑logged. With those tightened, this should noticeably improve day‑to‑day reliability.

If you want, I can draft inline GitHub review comments you can paste onto the PR (one per file/region) focusing on the specific lines in `PortManager.cs` and `ServerInstaller.cs`.

[1]: https://github.com/CoplayDev/unity-mcp/commit/32274a396586988addd6f7babc89509b826f17f6 "UnityMCP stability: robust auto-restart on compile/play transitions; … · CoplayDev/unity-mcp@32274a3 · GitHub"
[2]: https://github.com/CoplayDev/unity-mcp/commit/a65f10383ab5ec381adcbb6f8167f9e5c9d3a385 "feat(bridge): embed Python server into package and remove Git-based i… · CoplayDev/unity-mcp@a65f103 · GitHub"
[3]: https://github.com/CoplayDev/unity-mcp/commit/f24e124c1565036ec50d80007a43aa7e2daf112e "MCP: Embedded server reliability and UX\n\n- Embed-first installer: c… · CoplayDev/unity-mcp@f24e124 · GitHub"
[4]: https://github.com/CoplayDev/unity-mcp/commit/2f387d3417a7b85c901fa0f19e2aee64d35e35c3 "chore(ui): rename 'Re-Run Client Setup' to 'Bind to Clients' · CoplayDev/unity-mcp@2f387d3 · GitHub"
[5]: https://github.com/CoplayDev/unity-mcp/pull/192 "Enhancement: Stabilize MCP bridge: per‑project sticky ports, robust auto‑restart, UI fix by dsarno · Pull Request #192 · CoplayDev/unity-mcp · GitHub"
[6]: https://github.com/CoplayDev/unity-mcp/commit/f4bc7cd4fded943881c9e7d16271a802eb3ae400 "fix(ports): never hop to default when stored port busy; prefer stored… · CoplayDev/unity-mcp@f4bc7cd · GitHub"

msanatan:
Hey @dsarno, thanks for this! I haven't fully checked out everything, realistically by tomorrow I can be more thorough. That said, make a few adjustments?

- Merge into the `main` branch instead of `master`, which is there for backward compatibility. The `main` branch has some changes, so be sure to merge it into your branch in case of conflicts!
- Now that UnityMcpServer is part of the installed Unity package, can you change its name to `UnityMcpServer~`. Why? It's not actual plugin code, and the ~ tells Unity to ignore those files in the import process. That's good, changes to those files won't required asset DB refreshes: https://docs.unity3d.com/Manual/SpecialFolders.html
  - Once the folder name is changed, can you remove the .meta files under `UnityMcpServer~` as they're no longer needed?
- I don't think we need the "old" UnityMcpServer folder at the root level, so can you delete it? If we do need it, why not just merge everything into the new UnityMcpServer folder that's within the plugin?
- If the installation instructions have to change, update the `README`! But please merge main into your branch first to avoid the conflicts 😅 . If necessary update README-DEV as well


Screenshot 2025-08-09 at 12 02 48 PM

dsarno:
Adding the fixes from @msanatan, as well as a couple others, now.  Please wait to review until fixes are in, I will ping.

dsarno:
Per @msanatan 's requests
- Merge upstream/main: CoplayDev rebrand with bridge stability improvements
- Package Python server under UnityMcpServer~; remove redundant .meta files
- Minor README updates: switched "Auto‑Configure" → "Auto‑Setup" and clarified package cache hash path + new image of MCPEditorWindow

Other screw-tightening on domain reloading:
- Added stop‑before‑reload (OnBeforeAssemblyReload) + deferred init on editor idle → stable rebinds to the same per‑project port
- Breadcrumb logs for stop/start/ping help trace reloads; ping fast‑path kept clients responsive

Note: Observed one transient `read_console` miss (server reported it was down) during a reload; it recovered immediately once the bridge restarted itself on the same port (6400).  Can still make this slightly better UX-wise but shouldn't stall nearly as much as before.

Ready for review @msanatan @Scriptwonder .

tumml3r:
I'm successfully on this branch now after experiencing the issues discussed here: https://github.com/CoplayDev/unity-mcp/issues/195. Just noting here for posterity.

UnityMcpBridge/Editor/Helpers/PortManager.cs
```diff
@@ -31,15 +41,30 @@ public class PortConfig
...
+            // Try to load stored port first, but only if it's from the current project
+            var storedConfig = GetStoredPortConfig();
+            if (storedConfig != null && 
+                storedConfig.unity_port > 0 && 
+                string.Equals(storedConfig.project_path ?? string.Empty, Application.dataPath ?? string.Empty, StringComparison.OrdinalIgnoreCase) &&
+                IsPortAvailable(storedConfig.unity_port))
             {
-                Debug.Log($"Using stored port {storedPort}");
-                return storedPort;
+                if (IsDebugEnabled()) Debug.Log($"<b><color=#2EA3FF>UNITY-MCP</color></b>: Using stored port {storedConfig.unity_port} for current project");
```
msanatan, line 51: Note to self for later, set up a structured logger with levels and the like

UnityMcpBridge/Editor/Windows/UnityMcpEditorWindow.cs
```diff
@@ -1580,5 +2016,99 @@ private void CheckClaudeCodeConfiguration(McpClient mcpClient)
                 mcpClient.SetStatus(McpStatus.Error, e.Message);
             }
         }
+
+        private bool IsPythonDetected()
```
msanatan, line 2020: Note to self for later, check if there's a more generic way to do this

UnityMcpBridge/UnityMcpServer~/src/port_discovery.py
```diff
@@ -0,0 +1,155 @@
+"""
+Port discovery utility for Unity MCP Server.
+
+What changed and why:
+- Unity now writes a per-project port file named like
+  `~/.unity-mcp/unity-mcp-port-<hash>.json` to avoid projects overwriting
```
msanatan, line 6: Love this solution

UnityMcpBridge/UnityMcpServer~/src/tools/manage_gameobject.py
```diff
@@ -126,6 +127,10 @@ def manage_gameobject(
             # Use the get_unity_connection function to retrieve the active connection instance
             # Changed "MANAGE_GAMEOBJECT" to "manage_gameobject" to potentially match Unity expectation
             response = get_unity_connection().send_command("manage_gameobject", params)
+            if isinstance(response, dict) and not response.get("success", True) and response.get("state") == "reloading":
```
msanatan: Is there any non-connection reason why `response['success']` can be `False`? Also, why does it default to `True`? 

To be clear I'm not asking for it to be changed, just asking for my knowledge.

Non critical, future task could be to use a wrapper function around `get_unity_connection().send_command`

UnityMcpBridge/UnityMcpServer~/src/unity_connection.py
```diff
@@ -105,64 +109,133 @@ def receive_full_response(self, sock, buffer_size=config.buffer_size) -> bytes:
...
-            command_json = json.dumps(command, ensure_ascii=False)
-            self.sock.sendall(command_json.encode('utf-8'))
-            
-            response_data = self.receive_full_response(self.sock)
+            status = read_status_file()
+            if status and (status.get('reloading') or status.get('reason') == 'reloading'):
+                return {
+                    "success": False,
+                    "state": "reloading",
+                    "retry_after_ms": int(250),
```
msanatan: Why not use the value from `config.py`?

UnityMcpBridge/UnityMcpServer~/src/tools/manage_gameobject.py
dsarno: @msanatan 

> Is there any non-connection reason why response['success'] can be False? 

Yes, two cases:

(1) "Preflight" call finds Unity is reloading, not ready, so retry...
```
{"success": false, "state": "reloading", "retry_after_ms": 250}
```
(2) Tool ran but bad params or other type of fail
```
{"success": false, "error": "GameObject 'Foo' not found"}
```
(Preflight returns before the tool executes ... tool‑level errors only appear after preflight passes.)

> why does it default to True?

Overcautious and no longer strictly necessary, but ... the live ("old") version of the endpoint has no success field. So I kept a default to `true` in case we hit an older version of endpoint as we transition. In which case the missing `success` isn't treated as a fail, more like "this is the old version so go ahead"

Since server is now embedded, mixing old/new is unlikely; we could drop the default and do:

```
if isinstance(response, dict) and response.get("state") == "reloading" and response.get("success") is False:
    return response

```
But since people have been testing it and it seems to be working, maybe we slate that for a cleanup later with some of your other suggestions above.

UnityMcpBridge/UnityMcpServer~/src/unity_connection.py
dsarno: @msanatan Good call -- While addressing this, I found another issue with individual tools not retrying when they hit a recompile/reload gap.  Will work on it tomorrow and hopefully that can be the last bit for now. Other fixes can go in a follow-up.

dsarno: @msanatan Ok re your above note, I replaced the magic 250 with a single-sourced value in config.  Also realized individual tools weren't capable of retries after/during Unity reload, so added centralized reload-aware retries that every tool can use, so they retry instead of reporting fail and moving on.  You can see this behavior if you ask e.g. Cursor to run all 8 tool calls in a row, starting with a `manage_script` to force a recompile. 

Tools now:
- detect reload, sleep reload_retry_ms (default 250ms), retry up to reload_max_retries (default 40 ≈ 10s)
- preserve structured reloading failures if Unity takes longer
- use shared helpers, avoiding per-tool duplication

Follow-up possibilities (future PR)
- Add env overrides for reload_retry_ms and reload_max_retries
- Include a retry counter and total waited ms in timeout responses/logs
- Optionally expand reload detection to cover socket drops/exception paths
- Revisit defaults if we see reloads >10s in CI or large projects

UnityMcpBridge/UnityMcpServer~/src/tools/manage_gameobject.py
msanatan: Thanks for the detailed explanation!

UnityMcpBridge/UnityMcpServer~/src/unity_connection.py
msanatan: Awesome @dsarno, great catch and good remedy

msanatan:
Once merged this PR closes:

- #195 
- Maybe #190
- #181
- #103
- #54
- #92

Issue #199, opened 2025-08-10
Console read fails to retrieve error entries


ibytergj:
Having an issue with 

 “com.coplaydev.unity-mcp": "https://github.com/dsarno/unity-mcp.git?path=/UnityMcpBridge#feat/bridge-stability“

Where only messages and warnings are retrieved via the read console command - errors are not returned.

<img width="1197" height="309" alt="Image" src="https://github.com/user-attachments/assets/e33187d1-bfc5-4d2d-9844-c6e90ea20536" />

<img width="847" height="453" alt="Image" src="https://github.com/user-attachments/assets/076334fa-e608-4105-a6a7-bb27fa5fb4f0" />

<img width="1133" height="475" alt="Image" src="https://github.com/user-attachments/assets/829487ee-a499-45b8-8ba2-3e58081d741a" />

<img width="835" height="772" alt="Image" src="https://github.com/user-attachments/assets/eb47cbc8-48c2-4c57-abcd-17feba2fcd61" />

ibytergj:
So to follow up the issue appears be incomplete support for Assert Type error logs. If these types of errors could be retrieved from the Unity console that would be useful for complete log support.

            // Apply the observed "one level lower" correction
            switch (initialType)
            {
                case LogType.Error:
                    return LogType.Warning; // Error becomes Warning
                case LogType.Warning:
                    return LogType.Log; // Warning becomes Log
                case LogType.Assert:
                    return LogType.Assert; // Assert remains Assert (no lower level defined)
                case LogType.Log:
                    return LogType.Log; // Log remains Log
                default:
                    return LogType.Log; // Default fallback
            }

 """Gets messages from or clears the Unity Editor console.

        Args:
            ctx: The MCP context.
            action: Operation ('get' or 'clear').
            types: Message types to get **('error', 'warning', 'log', 'all').**
            count: Max messages to return.
            filter_text: Text filter for messages.
            since_timestamp: Get messages after this timestamp (ISO 8601).
            format: Output format ('plain', 'detailed', 'json').
            include_stacktrace: Include stack traces in output.

        Returns:
            Dictionary with results. For 'get', includes 'data' (messages).
        """

dsarno:
Duplicate of #125 .  Fix in progress.

dsarno:
Should be fixed by #202 .  Closing this issue, let us know if any problems.

PR #200, opened 2025-08-11
Fix: ReadConsole tool: stable log severity classification and reliable console reads across Unity versions

dsarno:
## Summary
This PR fixes inconsistent log severity in the `read_console` tool and makes console reads reliable across domain reloads and Unity versions. (See Issue https://github.com/CoplayDev/unity-mcp/issues/199).  Earlier version was frail; could miss Exceptions, misclassify errors, warnings and other log types.

## What changed
- Classify severity via stacktrace/message first (LogError/LogWarning/Exception/Assertion), with safe fallback to mode‑bit mapping
- Correct error/warning/log filtering; Exceptions/Asserts counted as errors for filtering
- Remove debug spam; tool now returns the current console buffer consistently
- Scope limited to `UnityMcpBridge/Editor/Tools/ReadConsole.cs` (no API/schema changes)

## Why
Unity’s internal `LogEntry.mode` bitfield can differ between versions, which caused misclassification (e.g., errors showing as warnings). Stacktrace‑first classification is stable and version‑resilient.

## How to verify
1. Generate one of each: `Debug.Log`, `Debug.LogWarning`, `Debug.LogError` (and optionally an exception/assert)
2. Run the `read_console` tool with type filters (`log`, `warning`, `error`) and confirm correct classification
3. Trigger a domain reload; `read_console` continues to return the current buffer accurately


dsarno:
My bad, need to wait till stability PR is merged, then this will be 1 commit.

PR #227, opened 2025-08-19
CI: add Claude NL suite workflow, prompt, and long Unity test script …

dsarno:
…on default branch for manual runs.  This will eventually become an auto-run suite of claude desktop prompts that will validates mcp behavior by talking to unity.


dsarno:
Dangit, accidentally PR'd this to the upstream repo again.

PR #229, opened 2025-08-20
Don't ignore the package's README

msanatan:
Fixes the bug found here: https://discord.com/channels/1330955894296543364/1407837016493789318/1407838019863449612


msanatan:
Ah, wrong files pushed up

PR #294, opened 2025-09-28
Fix the bug where the execution method fails to find the game object due to the missing `by_` prefix.

weberwang:
When I use Chinese prompts to execute an MCP task in VS Code, and search for a GameObject by name with the input:

```json
{
  "action": "find",
  "search_method": "name",
  "search_term": "ChessGame"
}
```

In the C# code, the `switch` matches `by_name`, and using other search methods will cause type issues.

### DIFF
--- a/UnityMcpBridge/Editor/Tools/ManageGameObject.cs
+++ b/UnityMcpBridge/Editor/Tools/ManageGameObject.cs
@@ -27,7 +27,7 @@ namespace MCPForUnity.Editor.Tools
             Converters = new List<JsonConverter>
             {
                 new Vector3Converter(),
-                new Vector2Converter(), 
+                new Vector2Converter(),
                 new QuaternionConverter(),
                 new ColorConverter(),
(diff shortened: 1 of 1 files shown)


msanatan:
Hey, thanks for submitting? Can you clarify this:

> and using other search methods will cause type issues.

Is it possible you can add a unit test to validate this bug and change? It doesn't have to be end-to-end, just a Unity test will suffice.

Also, would you have tried these changes using a large project i.e one with many scenes?

dsarno:
Oh I'm sorry, I didn't see this PR. I believe the one I just merged should fix this. Please try in latest version.

dsarno:
I only fixed the name search issue though, looks like you have some additional functionality that may still be good, if you want to test it with latest main changes.

weberwang:
After updating to the latest version, I haven’t encountered this issue recently. Thank you!

msanatan:
Good work @dsarno!

PR #296, opened 2025-09-30
fix: add parameter validation to manage_gameobject tool

dsarno:
Latest `manage_gameobject` had trouble with the "find" operation.  E.g. "`Find the "Player" object in the hierarchy`" returns "not found" even if object is there.  This fixes those issues.

- Prevent silent failures when using 'name' instead of 'search_term' for find action
- Add clear error messages guiding users to correct parameter usage
- Validate that 'search_term' is only used with 'find' action
- Update parameter annotations to clarify when each parameter should be used

Also added a `get_component` action to manage_gameobject, which returns data from a single component, rather than a full data dump of all serialized component data from `get_components` -- which should probably be deprecated come to think of it.

## Summary by CodeRabbit

- New Features
  - Constrained search method options for more predictable searches.
  - Runtime validation enforcing correct parameter combinations per action.

- Bug Fixes
  - Prevents using name for find actions; requires search_term when finding.
  - Blocks providing search_term for create/modify actions to avoid accidental misuse.

- Documentation
  - Clarified guidance: use name only for create/modify, and search_term only for find.
  - Improved help text for selecting a search method.

### DIFF
--- a/UnityMcpBridge/Editor/Tools/ManageGameObject.cs
+++ b/UnityMcpBridge/Editor/Tools/ManageGameObject.cs
@@ -155,6 +155,18 @@ namespace MCPForUnity.Editor.Tools
                             );
                         // Pass the includeNonPublicSerialized flag here
                         return GetComponentsFromTarget(getCompTarget, searchMethod, includeNonPublicSerialized);
+                    case "get_component":
+                        string getSingleCompTarget = targetToken?.ToString();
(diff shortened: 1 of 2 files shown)


dsarno:
This passes my claude test suite, and is relatively small change that fixes an annoying bug in manage_gameobject (in a feature I use a lot, :).   So I'm going to ram this through.

PR #348, opened 2025-10-24
Remove asset store download button

msanatan:
When the user does not install via Git, they're instead prompted to set up MCP servers with `uvx`. This can be the default moving forward, once I test the custom tools with this method

## Summary by CodeRabbit

* **New Features**
  * Added support for installing MCP server from Asset Store (remote path) alongside embedded local installation.
  * Introduced new MCP tools: apply_text_edits, script_apply_edits, and validate_script for enhanced capabilities.

* **Tests**
  * Expanded test coverage for remote server configuration paths.

* **Chores**
  * Updated project dependencies and Unity version requirements.
  * Simplified server management UI by removing the manual download option.

### DIFF
--- a/TestProjects/UnityMCPTests/Packages/manifest.json
+++ b/TestProjects/UnityMCPTests/Packages/manifest.json
@@ -1,6 +1,6 @@
 {
   "dependencies": {
-    "com.coplaydev.unity-mcp": "https://github.com/coplaydev/unity-mcp.git?path=MCPForUnity",
+    "com.coplaydev.unity-mcp": "file:../../../MCPForUnity",
     "com.unity.ai.navigation": "1.1.6",
     "com.unity.collab-proxy": "2.7.1",
     "com.unity.feature.development": "1.0.1",


TestProjects/UnityMCPTests/Packages/manifest.json
```diff
@@ -1,16 +1,16 @@
 {
   "dependencies": {
-    "com.coplaydev.unity-mcp": "https://github.com/coplaydev/unity-mcp.git?path=MCPForUnity",
-    "com.unity.ai.navigation": "1.1.4",
-    "com.unity.collab-proxy": "2.5.2",
+    "com.coplaydev.unity-mcp": "file:../../../MCPForUnity",
```
msanatan, line 3: @dsarno why was it changed to the Git URL? My test was failing and when I stepped through the code, it wasn't pointing to the new code I added in this PR. I then saw the manifest referenced the remote package.

This repo is how we run unit tests, I'm not sure how I can verify new tests locally if it's pointing to a remote Git package. Let me know if I'm missing something. We can always make more test projects for different use cases

TestProjects/UnityMCPTests/ProjectSettings/ProjectVersion.txt
```diff
@@ -1,2 +1,2 @@
-m_EditorVersion: 2022.3.6f1
-m_EditorVersionWithRevision: 2022.3.6f1 (b9e6e7e9fa2d)
+m_EditorVersion: 2022.3.62f2
```
msanatan, line 1: @dsarno sorry for not explaining some of the ideas with the test harness. I initially made the project with Unity 2021 because that was our minimum version supported. I'm not sure how many people still use so it may not be a big deal, but we lose a bit of coverage basing it on 2022. 

In any case, I've updated it to the 2022 version w/o the security bug. Though it primarily affects builds, I figure it's good to be on the safe side in case someone uses this project in ways we don't intend them to

MCPForUnity/Editor/Helpers/ConfigJsonBuilder.cs
```diff
@@ -40,19 +40,50 @@ public static JObject ApplyUnityServerToExistingConfig(JObject root, string uvPa
...
-            string effectiveDir = directory;
+            if (useRemote)
+            {
+                // Asset Store install - use remote uvx
+                string version = AssetPathUtility.GetPackageVersion();
+                string remoteUrl = $"git+https://github.com/CoplayDev/unity-mcp@v{version}#subdirectory=MCPForUnity/UnityMcpServer~/src";
+                
+                unity["command"] = "uvx";
+                unity["args"] = JArray.FromObject(new[] { "--from", remoteUrl, "mcp-for-unity" });
+            }
```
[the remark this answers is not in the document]
msanatan: This is only needed for Codex, a bug on Codex requires this env var. Other clients, so far, do not need it

dsarno:
@msanatan Apologies, those are mistakes. I had changed the manifest by accident during testing last night, and used my mcp-source script to change it back, forgetting that it needs to be relative. 

Unity version was not deliberate either. I did that while fixing some tests, and then didn't revert. You can change it all back.

msanatan:
> @msanatan Apologies, those are mistakes. I had changed the manifest by accident during testing last night, and used my mcp-source script to change it back, forgetting that it needs to be relative.
> 
> Unity version was not deliberate either. I did that while fixing some tests, and then didn't revert. You can change it all back.

No worries. You could have been doing an automated test runner or something, there are valid reasons for changing both, so I wanted to be sure I wasn't doing some of your work. I'll just leave as is if that's all right

dsarno:
> > @msanatan Apologies, those are mistakes. I had changed the manifest by accident during testing last night, and used my mcp-source script to change it back, forgetting that it needs to be relative.
> 
> > 
> 
> > Unity version was not deliberate either. I did that while fixing some tests, and then didn't revert. You can change it all back.
> 
> 
> 
> No worries. You could have been doing an automated test runner or something, there are valid reasons for changing both, so I wanted to be sure I wasn't doing some of your work. I'll just leave as is if that's all right 

@msanatan but you should definitely fix the manifest pathing. And at some point, we should go back to 2021, that was always a good idea. I can do it later if you like.

msanatan:
> > > @msanatan Apologies, those are mistakes. I had changed the manifest by accident during testing last night, and used my mcp-source script to change it back, forgetting that it needs to be relative.
> > 
> > 
> > > 
> > 
> > 
> > > Unity version was not deliberate either. I did that while fixing some tests, and then didn't revert. You can change it all back.
> > 
> > 
> > No worries. You could have been doing an automated test runner or something, there are valid reasons for changing both, so I wanted to be sure I wasn't doing some of your work. I'll just leave as is if that's all right
> 
> @msanatan but you should definitely fix the manifest pathing. And at some point, we should go back to 2021, that was always a good idea. I can do it later if you like.

Sure, but treat it low priority. I don't think Unity can downgrade version, so you may have to create a new one and copy the files

msanatan:
Superseeded by upcoming PR

PR #360, opened 2025-10-31
Add Multi-Instance Support for Unity MCP

sakura-habby:
# Add Multi-Instance Support for Unity MCP

## Overview
Enables MCP clients to interact with multiple Unity Editor instances simultaneously. Fully backward compatible with existing single-instance workflows.

### Core Features
- **Connection Pool**: Manages multiple Unity Editor connections automatically
- **Instance Discovery**: Auto-detects all running Unity instances from `~/.unity-mcp/`
- **Flexible Targeting**: Identify instances by project name, hash, or combined format
- **New Tool**: `list_unity_instances` - Lists all discovered Unity instances

### Multi-Instance Support Added To
- **All MCP Tools**: Now accept optional `unity_instance` parameter
- **All MCP Resources**: Support query parameters (e.g., `?unity_instance=ProjectA`)
- **Dual Resource Registration**: Each resource available in both fixed URI (default instance) and template URI (with parameters) formats

### Configuration
- Environment variable: `UNITY_MCP_DEFAULT_INSTANCE`
- CLI argument: `--default-instance`
- Auto-discovery fallback

## Backward Compatibility
✅ **Zero Breaking Changes** - All existing code works without modification. Multi-instance features are purely additive.

## Requirements
- Upgrade FastMCP to `>=2.13.0` for query parameter support

## Summary by CodeRabbit

* **New Features**
  * Multi-instance support: list running editors, per-instance targeting for tools/resources, CLI option to set a default instance, and a new "unity-instances" endpoint.
  * Connection pool for managing multiple editor connections; heartbeat/status now includes project name and Unity version.
  * Resource registry now exposes both instance-scoped and default resource variants.

* **Bug Fixes**
  * Improved port handling with automatic fallback when ports are busy and safer startup port switching.
  * Enhanced port discovery protocol with more robust probing and validation; cleanup of per-project status files at shutdown.

* **Chores**
  * Bumped fastmcp dependency to 2.13.0.

### DIFF
--- a/MCPForUnity/UnityMcpServer~/src/unity_connection.py
+++ b/MCPForUnity/UnityMcpServer~/src/unity_connection.py
@@ -225,7 +225,7 @@ class UnityConnection:
             logger.error(f"Error during receive: {str(e)}")
             raise
 
-    def send_command(self, command_type: str, params: dict[str, Any] | None = None) -> dict[str, Any] | MCPResponse:
+    def send_command(self, command_type: str, params: dict[str, Any] | None = None) -> dict[str, Any]:
         """Send a command with retry/backoff and port rediscovery. Pings only when requested."""
         # Defensive guard: catch empty/placeholder invocations early
         if not command_type:
@@ -276,12 +276,12 @@ class UnityConnection:
         try:
             status = read_status_file(target_hash)
             if status and (status.get('reloading') or status.get('reason') == 'reloading'):
-                return MCPResponse(
-                    success=False,
-                    error="Unity domain reload in progress, please try again shortly",
-                    data={"state": "reloading", "retry_after_ms": int(
-                        config.reload_retry_ms)}
-                )
+                return {
+                    "success": False,
+                    "state": "reloading",
+                    "retry_after_ms": int(config.reload_retry_ms),
+                    "error": "Unity domain reload in progress, please try again shortly",
+                }
         except Exception:
             pass
 
--- a/Server/unity_connection.py
+++ b/Server/unity_connection.py
@@ -225,7 +225,7 @@ class UnityConnection:
             logger.error(f"Error during receive: {str(e)}")
             raise
 
-    def send_command(self, command_type: str, params: dict[str, Any] | None = None) -> dict[str, Any] | MCPResponse:
+    def send_command(self, command_type: str, params: dict[str, Any] | None = None) -> dict[str, Any]:
         """Send a command with retry/backoff and port rediscovery. Pings only when requested."""
         # Defensive guard: catch empty/placeholder invocations early
         if not command_type:
@@ -276,12 +276,12 @@ class UnityConnection:
         try:
             status = read_status_file(target_hash)
             if status and (status.get('reloading') or status.get('reason') == 'reloading'):
-                return MCPResponse(
-                    success=False,
-                    error="Unity domain reload in progress, please try again shortly",
-                    data={"state": "reloading", "retry_after_ms": int(
-                        config.reload_retry_ms)}
-                )
+                return {
+                    "success": False,
+                    "state": "reloading",
+                    "retry_after_ms": int(config.reload_retry_ms),
+                    "error": "Unity domain reload in progress, please try again shortly",
+                }
         except Exception:
             pass
 


sakura-habby:
@coderabbitai generate docstrings

dsarno:
@sakura-habby Hi there.  I'm not able to get the `list-unity-instances` tool to appear in the MCP tool list, or your branch to recognize multiple instances.  Am I missing something?  

Is there anything else special that needs to happen to get the multi-instance functionality to work?  From the summary above it seems the ENV variable for default instance is optional, or is that necessary before this works?

image
image

sakura-habby:
@dsarno I didn't modify 'server_version.txt'. So maybe your "~/Library/Application\ Support/UnityMCP/UnityMcpServer/src/" hasn't been updated by Unity Plugin since they have the same version number?

dsarno:
@sakura-habby I’ve confirmed the tool works locally. Because we still have the server tools split between two locations, please add the new `list_unity_instances` module under `UnityMcpServer~/src/tools/` (where `register_all_tools` runs) and keep the `@mcp_for_unity_tool` decorator on the function so it’s picked up during startup.

cc @msanatan about the two server copies that we should probably clean up to avoid more confusion.

sakura-habby:
@dsarno What you said (sync 'MCPForUnity/UnityMcpServer~/src/' and 'Server/', use '@mcp_for_unity_tool') was already in my commits. However I didn't touch `UnityMcpBridge/` since it's an outdated leftover.

dsarno:
@sakura-habby This is understandably confusing, because we're running an experiment where we may take the server out of the package and fetch it with `uvx` (that's the `Server` directory). But the package currently embeds the entire server and copies it into the user's filesystem -- that's`MCPForUnity/UnityMcpServer~/src/`).  So yeah, we have two copies of the server and they should be exactly the same for now.  You're right you can ignore the deprecated `UnityMcpBridge/`, which is a third thing.

So for now, we still embed the python server in the package, and your new tool isn't in there.  If you just copy it in, it works when the user downloads the package -- otherwise, it doesn't get downloaded.  If you make that change and fix the uv.lock conflicts, we should be good to merge.

sakura-habby:
@dsarno On my branch, "MCPForUnity/UnityMcpServer~/src/" should be the same as "Server/". It's not copied automatically on your PC because "server_version.txt" is the same as existing one in '~/Library/...'. I think it should be updated by maintainer when releasing a new build, right?

dsarno:
@sakura-habby `server_version.txt` only controls when the Unity package copies UnityMcpServer~/src into `/Library/Application Support/UnityMCP/` at runtime. (In other words, "Installing" the server to the user's program directories).  It doesn’t propagate files between Server/ and UnityMcpServer~/src inside the repo.  Again, those are two separate copies of the server in the GitHub repo (not the Unity package), because we're experimenting with a new way of installing the server. One of those two will be removed from the repo when we decide whether to go the uvx route or keep the server embedded.

sakura-habby:
@dsarno I know what you mean. I had already done what you told me to do. I mean: when you test, you have to manually copy the files for now. That's because "server_version.txt" is not updated by me, not because "MCPForUnity/UnityMcpServer~/src/" is outdated.
BTW, I've updated uv.lock according to merged pyproject.toml, however there're still conflicts. What should I do?

dsarno:
@sakura-habby Check out the PR I submitted to your `multiple-instance` branch (this PR).  It should fix the uv.lock conflicts, and adds the `list_unity_instances` tool to the embedded server folder (`MCPForUnity/UnityMcpServer~/src/`).  I also fixed an issue where it was advertising "stale" instances of Unity that weren't actually live instances, so now it only looks at instances per port with the latest heartbeat.  

If you merge that, the merge conflicts here should go away, and we should be good to go.

Server/unity_connection.py
```diff
@@ -371,32 +392,249 @@ def read_status_file() -> dict | None:
                 raise
 
 
-# Global Unity connection
-_unity_connection = None
+# -----------------------------
+# Connection Pool for Multiple Unity Instances
+# -----------------------------
+
+class UnityConnectionPool:
```
msanatan: Nice, could be in a separate file to prevent even more bloating this file

Server/unity_connection.py
```diff
@@ -371,32 +392,249 @@ def read_status_file() -> dict | None:
...
+    """Manages connections to multiple Unity Editor instances"""
+
+    def __init__(self):
+        self._connections: Dict[str, UnityConnection] = {}
```
msanatan: `Dict[str, UnityConnection]` we don't really use this style of type formatting. LLMs typically do this by default for Python code, but this repo follows modern Python standards (the exceptions being the AI generated bits in this file and scripts)

`Optional[X]` -> `X | None`
`Dict[X, Y]` -> `dict[X, Y]`

Server/unity_connection.py
msanatan, line 1: I don't think I fully understand all of the networking code of this project, it's something I'm still getting in the weeds of, but I think there are some major logical flaws here.

This still reads the status file that was touched most recently, not the one for the targeted project. `read_status_file()` doesn't have a notion of `instance_id`, so if another Unity project for some reason has an updated status file (likely through a domain reload), I think that'll send commands to the wrong instance

Server/tools/list_unity_instances.py
```diff
@@ -0,0 +1,69 @@
...
+"""
+from typing import Annotated, Any
+
+from fastmcp import Context
+from registry import mcp_for_unity_tool
+from unity_connection import get_unity_connection_pool
+
+
+@mcp_for_unity_tool(description="List all running Unity Editor instances with their details.")
+def list_unity_instances(
```
msanatan: This is an MCP resource

msanatan (commented):
Hey @sakura-habby , thanks for submitting this PR! And sorry for taking so long, I've been caught up with some big changes in this project and others.

I do have an issue with polluting the function tools with an optional `unity_instance`. There's a chance that an LLM can hallucinate an ID - the IDs are meant to be deterministic.

Did you have a look at the FastMCP documentation? The `context` parameter allows for metadata to be added and read in the request, which I believe is a more suitable place to add the instance ID. Was that explored? 

Also check the potential timing issue with the status file, and how that mess up. This will be a blocker if I understood your code correctly

msanatan:
> @sakura-habby I’ve confirmed the tool works locally. Because we still have the server tools split between two locations, please add the new `list_unity_instances` module under `UnityMcpServer~/src/tools/` (where `register_all_tools` runs) and keep the `@mcp_for_unity_tool` decorator on the function so it’s picked up during startup.
> 
> cc @msanatan about the two server copies that we should probably clean up to avoid more confusion.

I think @sakura-habby did well to update Server, we'll be able to use that moving forward.

msanatan:
> @dsarno I know what you mean. I had already done what you told me to do. I mean: when you test, you have to manually copy the files for now. That's because "server_version.txt" is not updated by me, not because "MCPForUnity/UnityMcpServer~/src/" is outdated. BTW, I've updated uv.lock according to merged pyproject.toml, however there're still conflicts. What should I do?

I think I've gotten a way around this

Server/unity_connection.py
dsarno: > I don't think I fully understand all of the networking code of this project, it's something I'm still getting in the weeds of, but I think there are some major logical flaws here.
> 
> This still reads the status file that was touched most recently, not the one for the targeted project. `read_status_file()` doesn't have a notion of `instance_id`, so if another Unity project for some reason has an updated status file (likely through a domain reload), I think that'll send commands to the wrong instance

@msanatan What about this fix, where we use a hash to find the instance instead of the most recent?

https://github.com/sakura-habby/unity-mcp/pull/1/commits/198a4ffad2478a9520db867974da89948fb0a674

msanatan: @dsarno that should work!

msanatan:
> @sakura-habby Check out the PR I submitted to your `multiple-instance` branch (this PR). It should fix the uv.lock conflicts, and adds the `list_unity_instances` tool to the embedded server folder (`MCPForUnity/UnityMcpServer~/src/`). I also fixed an issue where it was advertising "stale" instances of Unity that weren't actually live instances, so now it only looks at instances per port with the latest heartbeat.
> 
> If you merge that, the merge conflicts here should go away, and we should be good to go.

Just had a look, great fixes!

sakura-habby:
@dsarno Ha sorry I didn't notice it. Thank you so much!

dsarno:
@msanatan Are you good to merge this? I can work on a separate patch to get the instance from context instead of having LLM fill it in.

dsarno:
> Hey @sakura-habby , thanks for submitting this PR! And sorry for taking so long, I've been caught up with some big changes in this project and others.
> 
> 
> 
> I do have an issue with polluting the function tools with an optional `unity_instance`. There's a chance that an LLM can hallucinate an ID - the IDs are meant to be deterministic.
> 
> 
> 
> Did you have a look at the FastMCP documentation? The `context` parameter allows for metadata to be added and read in the request, which I believe is a more suitable place to add the instance ID. Was that explored? 

@msanatan ~~I tried to get clear on this and from what I read, context is immutable -- you can't write to it or add custom parameters. But maybe we can store instance id data in another persistent structure across the session...?~~  Ok found newer info for get and set state -- looking into it.

MCPForUnity/UnityMcpServer~/src/unity_connection.py
```diff
@@ -371,32 +408,249 @@ def read_status_file() -> dict | None:
...
+            if self._default_instance_id:
+                instance_identifier = self._default_instance_id
+                logger.debug(f"Using default instance: {instance_identifier}")
+            else:
+                # Use the most recently active instance
+                # Instances with no heartbeat (None) should be sorted last (use epoch as sentinel)
+                from datetime import datetime
+                sorted_instances = sorted(instances, key=lambda i: i.last_heartbeat or datetime.fromtimestamp(0), reverse=True)
+                logger.info(f"No instance specified, using most recent: {sorted_instances[0].id}")
+                return sorted_instances[0]
```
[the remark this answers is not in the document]
sakura-habby: @coderabbitai This issue has been addressed during the refactoring.

The sorting logic was moved to `connection_pool.py:92-96` and fixed as suggested:

```python
sorted_instances = sorted(
    instances,
    key=lambda inst: inst.last_heartbeat.timestamp() if inst.last_heartbeat else 0.0,
    reverse=True,
)
```

The fix uses .timestamp() to convert to float and 0.0 as the sentinel value (instead of datetime.fromtimestamp(0)), avoiding the timezone-aware/naive datetime mixing issue.

sakura-habby:
> > Hey @sakura-habby , thanks for submitting this PR! And sorry for taking so long, I've been caught up with some big changes in this project and others.
> > I do have an issue with polluting the function tools with an optional `unity_instance`. There's a chance that an LLM can hallucinate an ID - the IDs are meant to be deterministic.
> > Did you have a look at the FastMCP documentation? The `context` parameter allows for metadata to be added and read in the request, which I believe is a more suitable place to add the instance ID. Was that explored?
> 
> @msanatan ~I tried to get clear on this and from what I read, context is immutable -- you can't write to it or add custom parameters. But maybe we can store instance id data in another persistent structure across the session...?~ Ok found newer info for get and set state -- looking into it.

@msanatan @dsarno Thank you both for the thoughtful feedback and for taking time to review this PR! I really appreciate the discussion about the design approach.

## Current Implementation: Stateless with Parameters

I chose the stateless approach (passing `unity_instance` as an optional parameter) based on my understanding of FastMCP's architecture and MCP protocol design. Here's my reasoning:

### Why Stateless?
1. **Aligns with MCP protocol philosophy**: Each tool call is self-contained with all necessary context
2. **Deterministic and explicit**: The LLM can query `unity_instances` resource to get valid instance IDs, reducing hallucination risk
3. **No external dependencies**: Doesn't require session storage or state management infrastructure
4. **Simpler lifecycle**: No concerns about state loss on server restart

### Why Not Context/Meta Approach?

I investigated both options mentioned:

**1. FastMCP Context (`ctx.set_state()` / `ctx.get_state()`)**
According to [FastMCP documentation](https://gofastmcp.com/servers/context):
> "Each MCP request receives a new context object, and context is scoped to a single request; state or data set in one request will not be available in subsequent requests."

This means:
- ✅ Works for middleware → tool within a **single request**
- ❌ Cannot persist across **multiple tool calls**
- ❌ Cannot implement "select instance, then remember it" pattern

**2. MCP Protocol `_meta` Field**
While the [MCP protocol spec](https://spec.modelcontextprotocol.io/) supports `_meta` in tool call requests, I couldn't find a clear mechanism for:
- How LLMs would set `_meta` values when calling tools
- How clients like Claude Desktop would expose this to the LLM
- Whether this is a standard pattern in production MCP servers

If `_meta` is a viable approach, I'd love to learn more about the intended usage!

### Alternative: Stateful with External Storage

To implement stateful instance selection (e.g., `select_unity_instance()` then subsequent calls remember it), we would need:

**Option A: Global Variable** (stdio transport only)
```python
_active_instance_per_session: dict[str, str] = {}  # session_id -> instance_id
```
- ✅ Works for stdio (process isolation)
- ❌ Risk of state loss on server restart
- ❌ Need to track session IDs

Option B: Persistent Storage
- Redis, SQLite, or file-based session store
- ✅ Survives restarts
- ❌ Introduces external dependency
- ❌ Adds complexity for session lifecycle management

Questions & Next Steps

I'm very open to exploring alternative approaches if there's a better way! Could you help clarify:

1. For context-based approach: Is there a pattern I'm missing for persisting state across tool calls in FastMCP?
2. For _meta approach: How would the LLM set metadata values? Is there documentation or examples of this pattern in production?
3. Design preference: Would you prefer I explore the stateful + external storage approach, or is there a third option I haven't considered?

BTW, although this reply is polished by Claude, I had reviewed it before it's posted.

dsarno:
@sakura-habby  thank you for all the effort you're putting in to this. 

One piece of feedback for you: I can tell you're relying a lot on models to help with your PR and coding-- that's fine, I do too. But you have to also thoroughly test your functionality to make sure it works. Your first submission didn't work because files weren't in the right place.  And then when I fixed that, I found that it was detecting stale Unity instances.  Then after more fixes, I found there are other issues with it being able to correctly target the instance it wants to send commands to. 

I recommend opening two or three live Unity instances and asking your tool to do a variety of mcp tool calls to each one (not just the list instances, which is fairly trivial) but actually sending mcp tool calls to the instance the user wants, and then switching to another one, especially across domain reloads. If you can get that working dependably, you're in better shape. Also I think the above analysis about storing values may not be correct. In playing with it last night, I was able to get the context to store instance id values across requests. 

So for example if you can get mcp to create a sphere and add a blue material in one instance, and then do the same in another one, all in one prompt, that's positive. 

Thanks again for all your work, We'd love to get this merged when it's working (and I haven't tried your most recent version yet so maybe it already works!)

sakura-habby:
@dsarno I'm terribly sorry about that. In fact I had tested with 3 different Unity projects before I sent this PR. And I'm already using it in my actual work without obvious issues. However for CodeRabbit feedbacks, I didn't test carefully and heavily reply on LLM with them. Sorry for wasting your time on it.

dsarno:
> @dsarno I'm terribly sorry about that. In fact I had tested with 3 different Unity projects before I sent this PR. And I'm already using it in my actual work without obvious issues. However for CodeRabbit feedbacks, I didn't test carefully and heavily reply on LLM with them. Sorry for wasting your time on it.

No need to apologize, we really appreciate your contribution. The repo is a bit complicated to develop for and test right now because of all the moving parts. I'm just offering a bit of guidance to help test. The main one is to make sure changes you make to your local repo are being copied into the live server directory (usually with rebuild server button), and that the Unity instances manifest.json are pointing to your local repo so they recompile if you change any of the Unity bridge / C# code. You may already be doing this but just in case.

msanatan:
> @msanatan Are you good to merge this? I can work on a separate patch to get the instance from context instead of having LLM fill it in.

I haven't tested it, if you're OK with it and think you can improve it, then sure.

msanatan:
> > Hey @sakura-habby , thanks for submitting this PR! And sorry for taking so long, I've been caught up with some big changes in this project and others.
> > I do have an issue with polluting the function tools with an optional `unity_instance`. There's a chance that an LLM can hallucinate an ID - the IDs are meant to be deterministic.
> > Did you have a look at the FastMCP documentation? The `context` parameter allows for metadata to be added and read in the request, which I believe is a more suitable place to add the instance ID. Was that explored?
> 
> @msanatan ~I tried to get clear on this and from what I read, context is immutable -- you can't write to it or add custom parameters. But maybe we can store instance id data in another persistent structure across the session...?~ Ok found newer info for get and set state -- looking into it.

Yea, set_state via Middleware. But @dsarno just explore it - I'm honestly not sure if it can work. I don't want to send you down a rabbit hole w/o a reason

msanatan:
> > > Hey @sakura-habby , thanks for submitting this PR! And sorry for taking so long, I've been caught up with some big changes in this project and others.
> > > I do have an issue with polluting the function tools with an optional `unity_instance`. There's a chance that an LLM can hallucinate an ID - the IDs are meant to be deterministic.
> > > Did you have a look at the FastMCP documentation? The `context` parameter allows for metadata to be added and read in the request, which I believe is a more suitable place to add the instance ID. Was that explored?
> > 
> > 
> > @msanatan ~I tried to get clear on this and from what I read, context is immutable -- you can't write to it or add custom parameters. But maybe we can store instance id data in another persistent structure across the session...?~ Ok found newer info for get and set state -- looking into it.
> 
> @msanatan @dsarno Thank you both for the thoughtful feedback and for taking time to review this PR! I really appreciate the discussion about the design approach.
> 
> ## Current Implementation: Stateless with Parameters
> I chose the stateless approach (passing `unity_instance` as an optional parameter) based on my understanding of FastMCP's architecture and MCP protocol design. Here's my reasoning:
> 
> ### Why Stateless?
> 1. **Aligns with MCP protocol philosophy**: Each tool call is self-contained with all necessary context
> 2. **Deterministic and explicit**: The LLM can query `unity_instances` resource to get valid instance IDs, reducing hallucination risk
> 3. **No external dependencies**: Doesn't require session storage or state management infrastructure
> 4. **Simpler lifecycle**: No concerns about state loss on server restart
> 
> ### Why Not Context/Meta Approach?
> I investigated both options mentioned:
> 
> **1. FastMCP Context (`ctx.set_state()` / `ctx.get_state()`)** According to [FastMCP documentation](https://gofastmcp.com/servers/context):
> 
> > "Each MCP request receives a new context object, and context is scoped to a single request; state or data set in one request will not be available in subsequent requests."
> 
> This means:
> 
> * ✅ Works for middleware → tool within a **single request**
> * ❌ Cannot persist across **multiple tool calls**
> * ❌ Cannot implement "select instance, then remember it" pattern
> 
> **2. MCP Protocol `_meta` Field** While the [MCP protocol spec](https://spec.modelcontextprotocol.io/) supports `_meta` in tool call requests, I couldn't find a clear mechanism for:
> 
> * How LLMs would set `_meta` values when calling tools
> * How clients like Claude Desktop would expose this to the LLM
> * Whether this is a standard pattern in production MCP servers
> 
> If `_meta` is a viable approach, I'd love to learn more about the intended usage!
> 
> ### Alternative: Stateful with External Storage
> To implement stateful instance selection (e.g., `select_unity_instance()` then subsequent calls remember it), we would need:
> 
> **Option A: Global Variable** (stdio transport only)
> 
> ```python
> _active_instance_per_session: dict[str, str] = {}  # session_id -> instance_id
> ```
> 
> * ✅ Works for stdio (process isolation)
> * ❌ Risk of state loss on server restart
> * ❌ Need to track session IDs
> 
> Option B: Persistent Storage
> 
> * Redis, SQLite, or file-based session store
> * ✅ Survives restarts
> * ❌ Introduces external dependency
> * ❌ Adds complexity for session lifecycle management
> 
> Questions & Next Steps
> 
> I'm very open to exploring alternative approaches if there's a better way! Could you help clarify:
> 
> 1. For context-based approach: Is there a pattern I'm missing for persisting state across tool calls in FastMCP?
> 2. For _meta approach: How would the LLM set metadata values? Is there documentation or examples of this pattern in production?
> 3. Design preference: Would you prefer I explore the stateful + external storage approach, or is there a third option I haven't considered?
> 
> BTW, although this reply is polished by Claude, I had reviewed it before it's posted.

Thanks for sharing. It is tricky, while FastMCP does support _meta properties, @dsarno alluded to it in another issue that MCP clients don't always support things that are defined in the spec. So using `_meta` might work for Claude and we'd have no idea what's happening in Cursor/Gemini. Not saying that's the case, but a lot of clients don't make it clear what they support from the spec, so these things happen more than we'd like

msanatan:
> @sakura-habby thank you for all the effort you're putting in to this.
> 
> One piece of feedback for you: I can tell you're relying a lot on models to help with your PR and coding-- that's fine, I do too. But you have to also thoroughly test your functionality to make sure it works. Your first submission didn't work because files weren't in the right place. And then when I fixed that, I found that it was detecting stale Unity instances. Then after more fixes, I found there are other issues with it being able to correctly target the instance it wants to send commands to.
> 
> I recommend opening two or three live Unity instances and asking your tool to do a variety of mcp tool calls to each one (not just the list instances, which is fairly trivial) but actually sending mcp tool calls to the instance the user wants, and then switching to another one, especially across domain reloads. If you can get that working dependably, you're in better shape. Also I think the above analysis about storing values may not be correct. In playing with it last night, I was able to get the context to store instance id values across requests.
> 
> So for example if you can get mcp to create a sphere and add a blue material in one instance, and then do the same in another one, all in one prompt, that's positive.
> 
> Thanks again for all your work, We'd love to get this merged when it's working (and I haven't tried your most recent version yet so maybe it already works!)

I'm really sorry @sakura-habby @dsarno, I should be able to test this later, I've just been caught up with a few different tasks at the same time.

One thing I'd like tested if you can get to it before me is verifying the connections work with a domain reload e.g. I have 2 unity instances open -> I've used MCP to make changes to the scene -> I write a script in one of the instances -> My MCP client has still make changes to the correct instance/scene.

Thanks for all your work @sakura-habby, and @dsarno for thoroughly reviewing and testing

dsarno:
> One thing I'd like tested if you can get to it before me is verifying the connections work with a domain reload e.g. I have 2 unity instances open -> I've used MCP to make changes to the scene -> I write a script in one of the instances -> My MCP client has still make changes to the correct instance/scene.

@msanatan I'll test the domain reloads after I play with the set_state  stuff. 

@sakura-habby Give me a day to work on it before we make any more changes. Then it's easier to stay in sync while I try a few things and test. Thanks!

sakura-habby:
> @sakura-habby thank you for all the effort you're putting in to this.
> 
> One piece of feedback for you: I can tell you're relying a lot on models to help with your PR and coding-- that's fine, I do too. But you have to also thoroughly test your functionality to make sure it works. Your first submission didn't work because files weren't in the right place. And then when I fixed that, I found that it was detecting stale Unity instances. Then after more fixes, I found there are other issues with it being able to correctly target the instance it wants to send commands to.
> 
> I recommend opening two or three live Unity instances and asking your tool to do a variety of mcp tool calls to each one (not just the list instances, which is fairly trivial) but actually sending mcp tool calls to the instance the user wants, and then switching to another one, especially across domain reloads. If you can get that working dependably, you're in better shape. Also I think the above analysis about storing values may not be correct. In playing with it last night, I was able to get the context to store instance id values across requests.
> 
> So for example if you can get mcp to create a sphere and add a blue material in one instance, and then do the same in another one, all in one prompt, that's positive.
> 
> Thanks again for all your work, We'd love to get this merged when it's working (and I haven't tried your most recent version yet so maybe it already works!)

@dsarno Oh sorry I've realised it's totally my fault. I have a global ".gitignore" that ignores "*\~", which matches "UnityMcpServer\~". So it works locally but I didn't commit it successfully with "git add -A .".

msanatan:
FYI @dsarno I'll trigger a small release in a bit, it shouldn't break anything here. Using `uvx` is working but I want to test more thoroughly before pushing that up for other reviews and testing, so that might go up for review today/tomorrow. We'll just ping whoever is done, so we can assist with any merge conflicts that arise

dsarno:
> FYI @dsarno I'll trigger a small release in a bit, it shouldn't break anything here. Using `uvx` is working but I want to test more thoroughly before pushing that up for other reviews and testing, so that might go up for review today/tomorrow. We'll just ping whoever is done, so we can assist with any merge conflicts that arise

@msanatan Sounds great. I might need another day anyway, handling some stuff on the homefront.

msanatan:
@sakura-habby are you on Discord/X by chance? We like highlighting contributors where possible

msanatan:
Replaced by #369

PR #365, opened 2025-11-03
It's time to let go, all dev for the plugin happens in MCPForUnity

msanatan:
True love is letting go CC @dsarno @Scriptwonder


dsarno:
> True love is letting go CC @dsarno @Scriptwonder

🤣🥲🫡

Thanks @msanatan

PR #371, opened 2025-11-10
[CUSTOM TOOLS] Roslyn Runtime Compilation Feature

Scriptwonder:
Add several custom tools that MCP Servers can call to dynamically generate code during any mode.

## Summary by CodeRabbit

* **New Features**
  * Runtime C# compilation and in-memory assembly loading with optional immediate execution during Play Mode.
  * Attach compiled MonoBehaviour scripts to GameObjects at runtime.
  * Editor window UI for composing, compiling, running, and inspecting code with diagnostics.
  * Commands/tools to compile, execute, list loaded assemblies/types, and manage compilation history.

* **Documentation**
  * Added Roslyn Runtime Compilation Tool documentation.

* **Bug Fixes**
  * Improved error reporting and diagnostics for compile/run failures.

### DIFF
--- a/CustomTools/RoslynRuntimeCompilation/ManageRuntimeCompilation.cs
+++ b/CustomTools/RoslynRuntimeCompilation/ManageRuntimeCompilation.cs
@@ -84,7 +84,10 @@ namespace MCPForUnity.Editor.Tools
             try
             {
                 string code = @params["code"]?.ToString();
-                string assemblyName = @params["assembly_name"]?.ToString() ?? $"DynamicAssembly_{DateTime.Now.Ticks}";
+                var assemblyToken = @params["assembly_name"];
+                string assemblyName = assemblyToken == null || string.IsNullOrWhiteSpace(assemblyToken.ToString())
+                    ? $"DynamicAssembly_{DateTime.Now.Ticks}"
+                    : assemblyToken.ToString().Trim();
                 string attachTo = @params["attach_to"]?.ToString();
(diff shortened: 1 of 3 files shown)


msanatan:
@Scriptwonder why is this a custom tool and not just a new tool?

Scriptwonder:
Mainly two reasons: 
(1) Generating and compiling code during play mode is not a request I see a lot of people have, so I think it is for my niche need, not to mention it's running on the Roslyn dependency, which some users have had issues with. I think of custoom tool as an optional toolbox that users can choose to download for enabling that feature (maybe in the future we can give them easier ways to download these optional features via remote server), which is kind of where this feature lies in.

(2) As you can see from the code, the runtime-generated code needs additional structure to store and save the code, which is additional overhead for the package.

I think Custom Tools are clean and descriptive ways to know what special features Unity-MCP have, with your underlying structure and tutorial makes it very easy to use. Is there concern for this decision? I think the migration between custom_tool and new tool is pretty easy so I can do it anytime, love to hear your thoughts on this.

msanatan:
I'd say put it in the main code, even though it's not a lot of people - it's more than you, hence why you're adding it in. But that's fair, if you think it's niche, then I guess it can be separate from the rest, all good!

I guess `PythonTools.asset.meta` wasn't meant to be uploaded? Maybe to `git rm --cached PythonTools.asset.meta` to remove it from the repo w/o removing it locally. I'm asking because it's the .meta file and not the asset!

PR #411, opened 2025-12-01
fix: restrict fastmcp version to avoid potential KeyError

Jordonh18:
This pull request introduces a minor update to the dependency specification for `fastmcp` in the `Server/pyproject.toml` file. The change restricts the allowed versions of `fastmcp` to exclude versions beyond `2.13.1`, avoiding the ctx keyerror

* Restricted the `fastmcp` dependency to versions `>=2.13.0,<2.13.2` in `Server/pyproject.toml` to prevent use of potentially incompatible newer versions.

### DIFF
--- a/Server/pyproject.toml
+++ b/Server/pyproject.toml
@@ -6,7 +6,7 @@ readme = "README.md"
 requires-python = ">=3.10"
 dependencies = [
  "httpx>=0.27.2",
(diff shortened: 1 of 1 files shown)


dsarno:
Thank you, I confirmed your diagnosis and tested the fix.  Merging now and will investigate updating unity-mcp to work with latest fastmcp update.

PR #413, opened 2025-12-02
Fix duplicate connection verification logs: add debounce and state-ch…

dsarno:
…ange-only logging

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved health-check handling to prevent overlapping verifications, reduce duplicate logs, and provide consistent health labels and UI indicators (Unknown/Healthy/Warning/Unhealthy).

* **Performance**
  * Added a debounce to global refreshes to avoid excessive refresh cycles and improve responsiveness.


dsarno:
Small cleanup of previous PR to reduce duplicate logs from toggline window (when debug logs is on)

PR #418, opened 2025-12-03
[FEATURE] Batch Commands

Scriptwonder:
Tested object generation/modification with batch, and it works perfectly! A feature we discussed a long time ago, but finally had an initial tryout at this.

**Description**: This is a new tool called **batch_execute** that can aggregate multiple commands together (to a max_number of 25 defined by both the Python server and C# side) and send them at once to achieve a more transparent and efficient (user-side) execution. In the tool description, LLM is explicitly told to work in a batch and sequence way to divide tasks and then process them sequentially.
image
Here is an example using Claude Desktop on Windows:
image

**Some notes:**
1: The current batching is a custom tool that tries to aggregate all the commands together and send them at once. It does not save tokens, and I doubt it will save many tokens even if we divide the batch into each tool (For example, have the manage_gameobject tool accept multiple parameters about multiple objects to do batch). It would definitely be useful when trying to perform redundant tasks, such as generating hundreds of similar pieces, but I also struggle to think of a way to save or optimize the token beyond the current scope.

2: I tried both VS Copilot and Claude Desktop. Claude Desktop works, but VS Copilot does not due to the nested structure of the batch. Will look into it more, but it is more likely to be a VS Copilot limitation based on previous integration issues back at July.

3: About Multi-thread: The original plan is to implement a multi-threaded option, but this seems to be limited to not only read-only tools, but also only a small portion of them (Multi-thread in Unity in my journey has always been a pain:( ). So I discard the changes of multi-thread and let it run on the main thread as of current moment. Happy to discuss about this.
image

CC'ing @msanatan and @dsarno here, let me know if you have any feedback!

## Summary by CodeRabbit

* **New Features**
  * Batch command execution tool to run up to 25 commands in a single request.
  * New Tools section in the editor with per-tool toggles, Built-in/Custom grouping, bulk Enable All / Disable All, Rescan, and a tools tab in the main window.

* **Improvements**
  * Accepts camelCase and snake_case parameter keys for better compatibility.
  * Tool enablement now persists and tools registered to clients reflect per-tool settings.

* **Removals**
  * Removed the Python-based Roslyn runtime compilation tool and its documentation.


Scriptwonder:
Will merge this to faciliate the gamejam tomorrow.

PR #420, opened 2025-12-03
[FIX] Temp Update on Material Assignment

Scriptwonder:
Previously it was shader assignment by request or iterative order from URP to Unlit, but LLM usually assign Standard in a URP setting.

This small fix introduce the helper to resolve some of the conflicts, while give warnings in the log to show if the user is creating a shader in a non-compatible setting.

## Summary by CodeRabbit

* **New Features**
  * Added centralized pipeline-aware shader resolution utility with automatic detection of Built-in, URP and HDRP pipelines and alias support.
* **Improvements**
  * Switched material handling to use the centralized shader resolver.
  * Improved messaging for shader compatibility and pipeline mismatches.
  * Material property application now strips shader-only entries and applies remaining properties only when present.


dsarno:
@Scriptwonder this is great and important. I was just thinking about tackling this yesterday. Thanks!

PR #434, opened 2025-12-05
Introduce Unified API Key Generation, Storage, and Authentication for MCP for Unity

Jordonh18:
This pull request introduces a unified API key management system for MCP for Unity, enabling secure authentication across server processes, HTTP, and WebSocket transports. The changes ensure the API key is generated, persisted, and surfaced in the UI, and update configuration and connection logic to require and utilize the key. The most important changes are grouped below.

**Authentication and API Key Management**

* Added new `AuthPreferencesUtility` class to handle API key generation, storage (EditorPrefs and disk), and retrieval, ensuring the API key is always available and consistent between client and server.
* Updated `EditorPrefKeys` to include a constant for storing the API key.
* Ensured the API key is created and persisted when launching the local HTTP server, so both the server and clients share the same key.
* WebSocket transport now sends the API key in both `X-API-Key` and `Authorization` headers during handshake, with improved error handling for authentication failures.

**User Interface Enhancements**

* Added UI components to display, copy, and regenerate the API key in the connection settings window (`McpConnectionSection.uxml` and code-behind), making it easy for users to manage their API key. [[1]](diffhunk://#diff-25bc852003c5b405393233c5d03e44d93235573f0ee7baf34f1806a5f9f56045R35-R37) [[2]](diffhunk://#diff-25bc852003c5b405393233c5d03e44d93235573f0ee7baf34f1806a5f9f56045R81-R83) [[3]](diffhunk://#diff-25bc852003c5b405393233c5d03e44d93235573f0ee7baf34f1806a5f9f56045R104-R111) [[4]](diffhunk://#diff-25bc852003c5b405393233c5d03e44d93235573f0ee7baf34f1806a5f9f56045R177-R200) [[5]](diffhunk://#diff-b43b2e07308fc754905c98d5cbc8bc09a17b3195b87de0df5818a8f86900af4dR14-R19) [[6]](diffhunk://#diff-25bc852003c5b405393233c5d03e44d93235573f0ee7baf34f1806a5f9f56045R267-R269) [[7]](diffhunk://#diff-25bc852003c5b405393233c5d03e44d93235573f0ee7baf34f1806a5f9f56045R541-R558)

**Configuration and Server Launch Updates**

* Updated manual config JSON builder logic to always require the API key for HTTP transport, removing legacy authorization and adding a prompt for the API key in the config. [[1]](diffhunk://#diff-b18fed24c75abf11bbf9d5bf02b5fe81f146718ed53f548c05a3c485ac7f4b51R18-R26) [[2]](diffhunk://#diff-b18fed24c75abf11bbf9d5bf02b5fe81f146718ed53f548c05a3c485ac7f4b51L38-R39) [[3]](diffhunk://#diff-b18fed24c75abf11bbf9d5bf02b5fe81f146718ed53f548c05a3c485ac7f4b51L51-R52) [[4]](diffhunk://#diff-b18fed24c75abf11bbf9d5bf02b5fe81f146718ed53f548c05a3c485ac7f4b51R66-R101)
* Improved server launch command construction to properly quote arguments and ensure the API key is available before starting the server. [[1]](diffhunk://#diff-df6c5b067a66aa03b72c528a85552e9e91b364f1109baf629c63e91ef9dc97e9L371-R389) [[2]](diffhunk://#diff-df6c5b067a66aa03b72c528a85552e9e91b364f1109baf629c63e91ef9dc97e9R533-R544)

**Meta and Miscellaneous**

* Added `.meta` files for new assets and folders to support Unity's asset tracking. [[1]](diffhunk://#diff-510680eaa97b3502d5a212394dd455e0ff3a8c18386c1351ec59dc38b6018017R1-R11) [[2]](diffhunk://#diff-923b80f54198ab08caf602386b073c4c7adf1f05f71b2a865d67f75f6da92b0eR1-R8)

These changes collectively provide a more secure and user-friendly authentication flow for MCP for Unity, ensuring all clients and servers use a shared API key and making key management straightforward for users.

### DIFF
--- a/MCPForUnity/Editor/Helpers/ConfigJsonBuilder.cs
+++ b/MCPForUnity/Editor/Helpers/ConfigJsonBuilder.cs
@@ -68,11 +68,6 @@ namespace MCPForUnity.Editor.Helpers
                 if (authEnabled)
                 {
                     var headers = unity["headers"] as JObject ?? new JObject();
-                    if (headers["X-API-Key"] != null)
-                    {
-                        headers.Remove("X-API-Key");
-                    }
-
                     EnsureInput(root, AuthTokenKey, AuthTokenDescription);
                     headers["Authorization"] = $"Bearer ${{input:{AuthTokenKey}}}";
                     unity["headers"] = headers;


dsarno:
@Jordonh18 Thanks for your contribution here.  This seems like a good idea for hardening the networking.  But it does add complexity and maintenance, and more room for bugs and potential security issues down the road.  So I guess my question is -- why now for this? Do you have a use case where you see more people wanting to run their own server remotely?  It seems the vast majority of users are running locally at this point.

Jordonh18:
The project already has extensive remote deployment infrastructure. The v8 migration explicitly redesigned the architecture to support remote scenarios. WebSocket communication replaced port 6400 specifically because a remote HTTP server wouldn't have access to a user's port 6400. The Docker Compose configuration binds to 0.0.0.0, and the documentation explicitly lists use cases like "run the MCP server in the cloud," "dedicated server all your personal computers connect to," and "hosted by various MCP marketplaces." These aren't theoretical features, they're documented capabilities of the current architecture.

The security gap is that these remote deployment features exist without any authentication mechanism. Right now, anyone deploying the Docker container or running the HTTP server on a network exposes full Unity Editor control to anyone who can reach that endpoint. This isn't hypothetical. The documented use cases (team servers, cloud hosting, marketplace distribution) all require access control. Without authentication, users must choose between using the remote features unsafely or not using them at all, which seems wasteful given the significant architectural work that went into enabling them.

A minimal API key implementation would be optional by default, preserving backward compatibility for local-only users. It would only activate when UNITY_MCP_API_KEY is set. This protects the users who are already deploying remotely (as your architecture enables and documents) without adding complexity for the majority running locally.

This is a long-term improvement, and I’m fully aware there are plenty of factors to weigh up. I’d really like to hear any concerns or arguments against it. I’m always open to feedback on anything I propose.

dsarno:
Hi, it makes sense, and I'm not against it in principle. In part this is a question of bandwidth (no pun intended). There are only a few active maintainers here, and I'm not a networking specialist -- so I need to look at it more closely,  even though I realize it's straightforward. We should get feedback from other maintainers like @msanatan and @Scriptwonder too.  

It's true the http upgrade was done in anticipation of a remote option, so that's a clear point in its favor -- or rather, a must-have for that use case. 

I guess if it ends up being easy enough to do this (meaning potential downsides, maintenance-costs and risks clearly understood by all), then it doesn't matter if most people won't use it yet, it should just be in there. 

I just want to be deliberate and get everyone's input before we go full network mode.

dsarno:
Hi again. I had a chance to look at this in more detail and now I see better how straightforward and simple it is.  I have no objections. 

If anything it will be making sure people know how to set it up -- but that will likely be advanced users anyway so they should be able to figure it out. I'll work on testing and merging this week. Thanks @Jordonh18 !

Jordonh18:
Ahh brilliant, so glad your onboard with the idea! Im doing some additional development to it today to patch up some holes and ensure its optional, as at the moment it cant be turned off. Need to implement the allow list as well, howver this one is less urgent as long as we have the token in there.

msanatan:
I've had some limited capacity to keep up with work and this repo recently @Jordonh18 @dsarno, I'll block out time tomorrow for a code review

Configurability wise, I'd love for this to be explicitly turned on via CLI args. Just so local users, our primary use case, don't have any additional work to do. From the convo it seems like that's the case.

Love the initiative!

Jordonh18:
Not been able to work on this recently but been back at it this evening. Now everything is enabled via CLI Flags, Its fairly clean. Unfortunately not every ide supports input:token in the mcp server config file. Not sure how you want to go about solving this issue and prompting for the token when the server starts.

i wanted to try and keep it as universal as possible and consistent across clients but just not possible. My next step would be maybe using env for these IDEs but wanted to stay away from that really as it may be too complicated for users.

Other than that it seems to work perfectly.

Any feedback or ideas would be great @dsarno @msanatan!

dsarno:
@Jordonh18 Sorry for the delay on this, and thanks again for the contribution -- we'd love to support continued contributions from you, as they've been great.

One idea on this one -- would it work to have the ENV set as part of the automatically generated config files that the MCP generates? That shouldn't be visible to users too much, and we could remove the ENV lines later as those non-supporting IDEs/clients support the tokens in config files.

dsarno:
@Jordonh18 Happy new year -- hope you're well.  Are you still interested in finishing this up? We'd love to tie it off, since we're actually going to put up a remote version of the server soon and this is a must-have for that.  Let us know!  

If so, please note the conflicts that have arisen since you first submitted (lots been happening in the last few weeks).

Jordonh18:
@dsarno Happy new year! Apologies been a very busy Christmas and even busier at work. I do indeed plan on finishing this very soon, i am working on it today. I like your ENV idea as a temporary solution, i'l implement and test!

msanatan:
The scope of this change is not 1:1 with the interoperable auth we set up in https://github.com/CoplayDev/unity-mcp/pull/644 so we'll close for now, however, there are elements we'd like to use from this PR moving onward. I do think a sample auth server might be useful for those hosting remotely

Issue #430, opened 2025-12-05
[ANTIGRAVITY BUG] Persistent "Context Canceled" / Client Disconnect despite Healthy Server Logs


IchBinChe:
Description
I am encountering a persistent connection issue between an MCP client (running within an AI Agent environment) and the mcp-for-unity-server. The client fails to establish a stable connection, reporting a "context canceled" error immediately after the handshake, while the server logs indicate that the requests are being received and processed successfully (200 OK / 202 Accepted).

Environment
OS: Windows
Server: mcp-for-unity-server (FastMCP 2.13.1)
Transport: HTTP (http://localhost:8080/mcp)
Client: Google Deepmind Antigravity Agent (Internal MCP Client)
Symptoms
Client-Side Error
The client repeatedly fails with the following error when attempting to call tools or list resources:

connection closed: calling "resources/list": client is closing: standalone SSE stream: failed to reconnect (session ID: <SESSION_ID>): connection failed after 5 attempts: Get "http://localhost:8080/mcp": context canceled
Note: The <SESSION_ID> changes after restarting the client/IDE, confirming that the client state is resetting, but the error persists with the new session.

Server-Side Logs
The server logs show that the connection is accepted and the initial handshake requests are successful. There are no errors on the server side.

[12/05/25 08:57:57] INFO     Starting MCP server 'mcp-for-unity-server' ...
...
2025-12-05 08:57:57,749 - mcp.server.streamable_http_manager - INFO - Created new transport with session ID: 257f88deefb74c59a42e319fa20e2cc6
INFO:     ::1:59388 - "POST /mcp HTTP/1.1" 200 OK
INFO:     ::1:59389 - "GET /mcp HTTP/1.1" 200 OK
INFO:     ::1:59390 - "POST /mcp HTTP/1.1" 202 Accepted
INFO:     ::1:59390 - "POST /mcp HTTP/1.1" 200 OK
Troubleshooting Steps Taken
Network Verification:

Confirmed TCP connectivity to localhost:8080 via PowerShell (Test-NetConnection) for both IPv4 (127.0.0.1) and IPv6 (::1).
Confirmed HTTP reachability via curl http://localhost:8080/mcp (returns 406 Not Acceptable as expected for non-SSE clients).
Firewall:

Verified that Python/Uvicorn is allowed through the Windows Firewall for both Private and Public networks.
Client Reset:

Restarted the client IDE to clear any stale session state.
[...]

PR #457, opened 2025-12-13
Fix Error: calling "initialize": invalid trailing data at the end of stream.

gajzzs:
https://github.com/CoplayDev/unity-mcp/issues/430

## Summary by CodeRabbit

* **New Features**
  * Added an output sanitizer that normalizes line endings and removes stray carriage returns from console output on Windows, improving console readability.
* **Bug Fixes**
  * Improved stdout handling during stdio operations to ensure consistent LF newlines and prevent unwanted formatting characters in console output, reducing display glitches.

### DIFF
--- a/Server/src/utils/cr_stripper.py
+++ b/Server/src/utils/cr_stripper.py
@@ -1,18 +1,75 @@
+"""
+Provides a utility to strip carriage return characters from output streams.
 
-# This utility strips carriage return characters (\r) from bytes or strings 
+This module implements the `CRStripper` class, which wraps a file-like object
+to filter out carriage return (\r) characters during write operations.
+
+Usage of this wrapper is essential for Model Context Protocol (MCP) communication
+over stdio, as it ensures consistent line endings and safeguards against 
+protocol errors, particularly in Windows environments.
+"""
+
+from typing import Any, BinaryIO
+
 class CRStripper:
-    def __init__(self, stream):
+    """
+    A file-like wrapper that strips carriage return (\r) characters from data before writing.
+    
+    This class intercepts write calls to the underlying stream and removes all 
+    instances of '\r', ensuring that output is clean and consistent across 
+    different platforms.
+    """
+    def __init__(self, stream: BinaryIO) -> None:
+        """
+        Initialize the stripper with an underlying stream.
+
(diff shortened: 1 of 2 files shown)


gajzzs:
@dsarno is there any issue in this PR ?

msanatan:
Thanks for your help with this @gajzzs, from discussions on the issue and this PR.

Can you help me fully understand your changes and answer these questions?

```python
msvcrt.setmode(sys.stdin.fileno(), os.O_BINARY)
```

Why does `stdin` need to change as well?

Wouldn't this only need to be applied for stdio mode, and not HTTP mode?

Is it possible to apply this fix w/o globally rewriting `sys.stdout`

Can you do the byte replacement on `sys.stdout.buffer`, and not the entire `sys.stdout` object?

For this fix to work, we need to patch stdout before running `mcp.run` right? If that's true, isn't it better to add this code to `utils` as a new Python module, and then import it to `main.py`? It looks quite messy in it's current location

Server/src/utils/cr_stripper.py
```diff
@@ -0,0 +1,18 @@
+
+# This utility strips carriage return characters (\r) from bytes or strings 
+class CRStripper:
```
msanatan, line 14: Can you:

- Remove the empty newline at the top
- Replace the Python comment with an actual docstring for the class
- Add type hints for `stream`, `data` and `name` parameters
  - If there are more than one type, use `|` instead of `Optional`/`Union` please e.g. `int | None` instead of `Optional[int]`

msanatan:
Looking good @gajzzs. I left one more comment style wise. The only thing pending after those changes would be other people testing, just to ensure it doesn't break other clients. I'll ask some in the community

dsarno:
Hi, I'm going to close this for now since, as far as I can tell, Antigravity now works in Windows for the MCP, so this shouldn't be necessary.  Holler if anyone disagrees.

msanatan:
Just to add to @dsarno's message, thanks @gajzzs but in a previous release we noticed Antigravity has been working since they updated their IDE. It's safer to have users update their IDE than to make this change to how we format stdout. 

We appreciate your efforts, thanks

Issue #474, opened 2025-12-20
Tools are often run without first ensuring the editor has updated the asset db and/or compiled


strich:
Often code is changed and the LLM will request a console read back or try to run tests, but the editor hasn't first been told to refresh the asset db and recompile. It doesn't do this automatically because it doesn't have focus. All tool calls should ensure the editor has refreshed and the calls should survive a recompile.

dsarno:
Can you offer a repro case? There are only a few tools that change code (manage_script basically).  And they reload domain reliably. 

If you're talking about your IDE (Claude, cursor) making changes directly, that doesn't force a reload, but Unity will reload on next focus.

strich:
I am indeed talking about using an external IDE. It's not viable to have to manually pull focus on the editor every time the AI wants to do something with it - Often the AI may be spending many minutes writing code, reading from the console and fixing bugs. Having to check back manually to pull focus then allow the read_console tool is frustrating and a time sink - That kind of tool call should happen in the background without my intervention.

A good solution here would be to provide an option in the editor window settings of the package to run `AssetDatabase.Refresh()` on every tool call. This is assuming tool calls survive recompilations?

dsarno:
Yes, that's less a limitation of the mcp than the way IDEs work, but I agree it'd be nice to give the IDE a tool to reload domain when it wants to check its work. (This is how Coplay works). You would only want to call it on code changes because otherwise you don't need a recompile / domain reload.

strich:
I disagree - it's not up to the IDE to tell unity things have changed. It's a failing of the MCP tools to ensure it's operating on the latest version of the assets. It would be silly to expect every IDE to somehow know it's working with a unity project and then also somehow know how to talk to the editor right?
Also you do want to call this on all tools all the time - calling assetdatabase.refesh() won't domain reload or recompile if no code has changed. It is a harmless call to make all the time

dsarno:
Can't disagree with a well stated case. Thanks for pushing back. 

Don't suppose you'd like to contribute a PR for it?

strich:
The only thing I'm concerned about is whether you've made the tool calls durable over editor recompilations? I have no idea.

dsarno:
Yes tools will retry if editor is compiling, then execute when editor is ready.  

See e.g.: https://github.com/CoplayDev/unity-mcp/blob/main/MCPForUnity/Editor/Services/HttpBridgeReloadHandler.cs

[...]

PR #475, opened 2025-12-22
Added AssetDatabase.Refresh() to all command tool calls

strich:
Per issue: https://github.com/CoplayDev/unity-mcp/issues/474
Added AssetDatabase.Refresh() to all command tool calls to ensure that the tools are working with the latest version of the editor assets. This should allow the editor to recompile and update assets in the background without manual intervention.

## Summary by Sourcery

Bug Fixes:
- Force an AssetDatabase refresh before executing registered editor commands to avoid operating on stale assets.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved command execution reliability by ensuring the asset database is automatically synchronized before executing all registered commands, both synchronously and asynchronously, to maintain consistency and prevent potential asset-related synchronization issues.

### DIFF
--- a/MCPForUnity/Editor/Tools/CommandRegistry.cs
+++ b/MCPForUnity/Editor/Tools/CommandRegistry.cs
@@ -8,6 +8,7 @@ using MCPForUnity.Editor.Helpers;
 using MCPForUnity.Editor.Resources;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
+using UnityEditor;
 
 namespace MCPForUnity.Editor.Tools
 {
@@ -232,6 +233,7 @@ namespace MCPForUnity.Editor.Tools
         /// <returns>The result for synchronous commands, or null for async commands (TCS will be completed later)</returns>
         public static object ExecuteCommand(string commandName, JObject @params, TaskCompletionSource<string> tcs)
         {
+            AssetDatabase.Refresh();
             var handlerInfo = GetHandlerInfo(commandName);
 
             if (handlerInfo.IsAsync)
@@ -256,6 +258,7 @@ namespace MCPForUnity.Editor.Tools
         /// <param name="params">Parameters to pass to the command (optional).</param>
         public static Task<object> InvokeCommandAsync(string commandName, JObject @params)
         {
+            AssetDatabase.Refresh();
             var handlerInfo = GetHandlerInfo(commandName);
             var payload = @params ?? new JObject();
 


strich:
It is worth noting that in my limited testing, tools like `run_tests` still report a time out during a recompile event. I think some additional resiliency is required @dsarno as I don't believe it was a genuine time out due to lack of time.

dsarno:
@strich Thanks for getting this going. But This was one of my concerns that the AI review has flagged. The asset db refresh command is not cheap or fast. So we should smarten it up to only happen when needed, not every time. I believe it's only needed when we touch code, which many of the tools don't do.

strich:
> @strich Thanks for getting this going. But This was one of my concerns that the AI review has flagged. The asset db refresh command is not cheap or fast. So we should smarten it up to only happen when needed, not every time. I believe it's only needed when we touch code, which many of the tools don't do.

The AI is wrong though. I have multiple massive commercial projects - 30GB+ with complex custom importers etc and calling `AssetDatabase.Refresh()` takes literally 0 time because it only refreshes dirty and new assets. The LLM is just wrong mate. :)

I think it is also wrong to suggest improving the batch command call - If the batching makes successive calls to the editor and some of those calls are making changes, then yes we should be calling refresh after each. If that certainly not happening, then sure it could be removed. But let's also remember that this call is harmless when there have been no changes so AFAIK there is no downside to calling it often.

dsarno:
Ok, assuming that's true, have you stress tested it?

strich:
> Ok, assuming that's true, have you stress tested it?

Not at all. I will try to do so with my real project over the next week. However as said earlier, the run tests tool call times out through a recompile. Would be good to get your input on that.

dsarno:
Thanks that would be great. 

I didn't write the run tests tool. I believe that was @JohanHoltby -- any idea why it would timeout Johan?

JohanHoltby:
@dsarno no I have not written this tool. Mine is an external test tool I sell on the assetstore - MooseRunner. All this relates to my other issue 416. LLMs get all jittery when we run the long recompiles. And yes there is a lot of state changes around it that can screw us over and if user uses hotreload we get more complexity. :). 

But yes never set it to run always. That will be super slow. And there are parameters in Unity on how recompile should be checked and different levels. I only do the assetrefresh when recompile fail.

(I have rolled back to 7.0.0 (super stable -1 install bug) as I need to develop a bunch of features as I'm on a strict deadline and the newer versions is not as stable and is missing 416 to be usable for me and i don't have the time to fix the problem. :(
)

Happy holidays

strich:
> LLMs get all jittery when we run the long recompiles. And yes there is a lot of state changes around it that can screw us over and if user uses hotreload we get more complexity. :).

This to me is the real problem with this MCP - It seems really fragile to editor refresh/recompile, which is absolutely critical to working with it. It's really beyond my loose understanding of the repo to try to improve it at this stage.
@dsarno it would be good to see some work happen on this - It's not really at the tool implementation level either. This sort of change should be lower level IMO.

> But yes never set it to run always. That will be super slow. And there are parameters in Unity on how recompile should be checked and different levels. I only do the assetrefresh when recompile fail.

Can you validate that opinion further? Under what circumstances is `AssetDatabase.Refresh()` slow when it shouldn't be? That is to say, it would only be slow when there are changes to import which needs to happen anyway. 
I guess what I'm arriving at here is - Under what circumstances should the MCP tools and the calling LLM _ever_ want to be operating on old outdated assets states? As far as I'm concerned this should never be the case, and waiting for the editor to update its state should be the default.
I could definitely understand a call to make this feature opt-out for some reason, but right now I can't see why you would.

JohanHoltby:
@strich 
> This to me is the real problem with this MCP - It seems really fragile to editor refresh/recompile, which is absolutely critical to working with it. It's really beyond my loose understanding of the repo to try to improve it at this stage.

I check if recompile needs to be done if so I recompile and no errors. If it recompiles and find errors then it throws errors. So this all is a lot of state transitions.
I only do the asset refresh as a command that the AI can do if i remember right.

In my testing i remember that when I do asset refresh it always do a domain reload.

But yes AI should not be forced to do a lot of checking. Just one command to run one method, class or assembly. - this saves context and reduce confusion for AI.

strich:
> In my testing i remember that when I do asset refresh it always do a domain reload.

Indeed no - It will not force a domain reload unless a script has changed. So its free so long as nothing changed.

JohanHoltby:
Great then! :) I use hotreload, maybe that was why?? I remember I had a lot of problems just setting it to always.... I had same argument.

msanatan:
If I add game objects to the scene, this unnecessarily refreshes assets. I don't think this is a good approach, it should be conditional for activities that need it

Scriptwonder:
> Great then! :) I use hotreload, maybe that was why?? I remember I had a lot of problems just setting it to always.... I had same argument.

How effective do you think introducing hotreload to the current tools could bring considering the effort? Currently in Play Mode doing some CRUD operations would still trigger the scene reset which is very inconvenient.

JohanHoltby:
@Scriptwonder its great for later steps when AI is finetuning the tests but worthless for first iteration:). So it has it place and saves time for me. I guess about 50%. I aim to do some videos after Christmas but they need a lot of editing. Prompting and waiting 20 min and then trigger research and then 10 min later trigger some new iteration which takes another 30 min is dead boring for most part... :)

dsarno:
Hi I'm going to close this for now. We think a global and ubiquitous db refresh call is a bit too blunt an instrument for this issue.  A better approach would be a tool the LLM can call to trigger a recompile/domain reload, preventing user from needing to switch focus to unity when the IDE touches any code.

strich:
Happy new year everyone.

> If I add game objects to the scene, this unnecessarily refreshes assets. I don't think this is a good approach, it should be conditional for activities that need it

If you mean it'll call Refresh() - yes it will. But it won't actually do anything if all the LLM did was add a game object to the scene. So I'm not sure what your concern is.

> Hi I'm going to close this for now. We think a global and ubiquitous db refresh call is a bit too blunt an instrument for this issue. A better approach would be a tool the LLM can call to trigger a recompile/domain reload, preventing user from needing to switch focus to unity when the IDE touches any code.

Fair enough re this PR - I accept it's a major change without any opt-out. However I think your proposed solution is not acceptable - Trying to get an LLM to remember to tool-call a refresh action before every other action is not a solution that is going to work I'm afraid. 
As it stands right now its just plain unusable for most users without the Refresh(). Maybe you're seeing this from the POV of Coplay environment? Most people using this MCP in any other standard IDE environment setup is going to have trouble as I have had. 

What really needs to happen is that this should be default on and an opt-out option for the few people who have edge-cases where its a problem for them. It could probably also be implemented with a little more finesse re the batch call etc.

PR #494, opened 2025-12-29
Optimise so startup is fast again

msanatan:
This PR does a few things:

- Lazy load tools in the UI
- Find tools by looking for the `McpForUnityTool` attribute via reflection
- We detect built-in tools by checking if their namespace is `MCPForUnity.Editor.Tools` instead of checking the asset path, which took a while

Startup time went from 6-12 seconds to around 200-500ms.

## Release Notes

* **Refactor**
  * Optimized tool discovery mechanism for improved performance when loading tools.
  * Implemented lazy loading for the Tools section in the editor window, reducing initial load time.
  * Simplified built-in tool detection logic.

* **Chores**
  * Cleaned up unused asset metadata files.

### DIFF
--- a/TestProjects/UnityMCPTests/Assets/Temp.meta
+++ b/TestProjects/UnityMCPTests/Assets/Temp.meta
@@ -1,8 +1,0 @@
-fileFormatVersion: 2
-guid: 20332651bb6f64cadb92cf3c6d68bed5
-folderAsset: yes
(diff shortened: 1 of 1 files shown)


dsarno:
👏👏👏

msanatan:
Did some testing, I think users will appreciate this one!

PR #521, opened 2026-01-07
Asset store helper script + updated README

msanatan:
I also replaced all the Debug.Log calls


msanatan:
The changes are huge because I added another Unity project in the repo for asset store uploads

PR #542, opened 2026-01-10
Use localhost to ping server if server binds to 0.0.0.0

kripergvg:
Problem: When the MCP server is configured to bind to 0.0.0.0:8080 (I am doing it locally to be able to access MCP from my local windows and from WSL), the WebSocket client would attempt to connect to 0.0.0.0, which fails because 0.0.0.0 is only valid for server binding, not client connections.

Solution: Added automatic URL translation that replaces 0.0.0.0 with localhost before establishing the WebSocket connection. This allows the server to bind on all interfaces while ensuring the client connects to the correct loopback address.

## Summary by Sourcery

Bug Fixes:
- Translate 0.0.0.0 to localhost in the constructed WebSocket endpoint URL to prevent client connection failures when the server listens on all interfaces.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebSocket endpoint construction and host normalization to handle edge host addresses (e.g., 0.0.0.0 → localhost).
  * Ensures the correct WebSocket scheme (wss/ws) is selected based on the connection protocol.
  * Consistently appends the required hub path segment so connections use a well-formed endpoint.

### DIFF
--- a/MCPForUnity/Editor/Services/Transport/Transports/WebSocketTransportClient.cs
+++ b/MCPForUnity/Editor/Services/Transport/Transports/WebSocketTransportClient.cs
@@ -682,16 +682,18 @@ namespace MCPForUnity.Editor.Services.Transport.Transports
                 throw new InvalidOperationException($"Invalid MCP base URL: {baseUrl}");
             }
 
-            string scheme = httpUri.Scheme.Equals("https", StringComparison.OrdinalIgnoreCase) ? "wss" : "ws";
-            string builder = $"{scheme}://{httpUri.Authority}";
-            if (!string.IsNullOrEmpty(httpUri.AbsolutePath) && httpUri.AbsolutePath != "/")
+            // Replace 0.0.0.0 with localhost for client connections
+            // 0.0.0.0 is only valid for server binding, not client connections
+            string host = httpUri.Host == "0.0.0.0" ? "localhost" : httpUri.Host;
+
+            var builder = new UriBuilder(httpUri)
             {
-                builder += httpUri.AbsolutePath.TrimEnd('/');
-            }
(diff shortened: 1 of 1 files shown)


msanatan:
Thanks @kripergvg! The code looks fine. Just so I can understand the problem right now:

1. You use the `uvx` command to start the server in WSL (as start local only fires on the host machine)
2. In the plugin, you enter `http://localhost:8080` as the URL
3. You press "Start Session"
4. The websocket server doesn't connect

Is that what happens now?

kripergvg:
> Thanks @kripergvg! The code looks fine. Just so I can understand the problem right now:
> 
> 1. You use the `uvx` command to start the server in WSL (as start local only fires on the host machine)
> 2. In the plugin, you enter `http://localhost:8080` as the URL
> 3. You press "Start Session"
> 4. The websocket server doesn't connect
> 
> Is that what happens now?

Thanks for reviwing the PR. 
Not exactly, 
1) I use uvx command to start the server in Windows
2) In the plugin, I enter http://0.0.0.0:8080 as the URL
3) I press start
4) Websocket doesn't connect because it tries to establish a connection wtih 0.0.0.0 instead of localhost.

0.0.0.0 is supposed to listen to all available network interfaces. I am using this in order to be able to use unity-mcp from both: windows and inside wsl

fgfg54321:
Yes, I also encountered the same problem. I hope it can be resolved as soon as possible.

msanatan:
Sorry for the delay @kripergvg, life required a lot of my attention elsewhere.

Thanks for clarifying, and making this change!

If you don't mind, share your Discord/LinkedIn/Twitter profiles with me and I'll tag you in the next release notes.

PR #559, opened 2026-01-15
fix: Prevent infinite compilation loop in Unity 6 when using wait_for_ready

### LINKED ISSUE #555: Recompile/Refresh occuring multiple times in sequence.
So I started using claude code recently instead of gpt. And I have been running into a thing where whenever It recomiles the editor so it can check for errors the editor refreshes/recompiles multiple times in sequence. I even gave it explicit instructions to only call once and wait for x seconds before looking at the console and its still happening. Is this a known/intended thing or is this an issue on my end? I'm really struggling to work it out.

BlindsidedGames:
## Summary

Fixes #555  - Infinite compilation/refresh loop when using `refresh_unity` with `compile="request"` and `wait_for_ready=true` in Unity 6.

## Root Cause

In Unity 6, the `WaitForUnityReadyAsync` method's use of `EditorApplication.update += Tick` to poll for compilation completion causes infinite domain reload loops. The async Task + EditorApplication.update callback combination doesn't survive domain reloads properly, causing the editor to repeatedly recompile until timeout.

## Fix

Skip `WaitForUnityReadyAsync` when `compileRequested` is true. Return immediately and let the client poll `editor_state` resource instead.

This is a minimal, targeted fix that:
- Only affects the specific case of `compile="request"` + `wait_for_ready=true`
- Maintains backward compatibility for all other refresh scenarios
- Follows the existing pattern of returning `resulting_state="compiling"` with a hint to poll

## Testing

**Environment:**
- Unity 6000.3.0f1 (Unity 6)
- MCP for Unity v9.0.7
- Claude Code VS Code Extension

**Before fix:** `refresh_unity` with `compile="request"` and `wait_for_ready=true` causes ~3 compilation loops before timing out.

**After fix:** Single compilation, immediate return with `resulting_state="compiling"`.

## Notes

- The `QueuePlayerLoopUpdate` changes in #558 were **not** the cause of the loop
- The issue is specifically the `EditorApplication.update` callback behavior across domain reloads in Unity 6

🤖 Generated with [Claude Code](https://claude.ai/code)

## Summary by Sourcery

Prevent infinite compilation loops in Unity 6 when using refresh commands that request compilation and wait for readiness.

Bug Fixes:
- Avoid infinite compilation/refresh loops in Unity 6 by skipping the wait-for-ready flow when a compile has been explicitly requested and letting clients poll editor state instead.

Enhancements:
- Gate the wait-for-ready behavior on a new conditional flag that preserves existing behavior on earlier Unity versions while applying the fix only on Unity 6+.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved editor refresh logic to avoid unnecessary waiting during compilation or domain reloads, improving responsiveness.
  * Updated post-refresh messaging so guidance accurately reflects whether the editor actually waited or is polling, including behavior adjustments for newer Unity versions.

### DIFF
--- a/MCPForUnity/Editor/Tools/RefreshUnity.cs
+++ b/MCPForUnity/Editor/Tools/RefreshUnity.cs
@@ -90,12 +90,17 @@ namespace MCPForUnity.Editor.Tools
                 return new ErrorResponse($"refresh_failed: {ex.Message}");
             }
 
-            // Unity 6 fix: Skip wait_for_ready when compile was requested.
+            // Unity 6+ fix: Skip wait_for_ready when compile was requested.
             // The EditorApplication.update polling in WaitForUnityReadyAsync doesn't survive
-            // domain reloads properly in Unity 6, causing infinite compilation loops.
+            // domain reloads properly in Unity 6+, causing infinite compilation loops.
             // When compilation is requested, return immediately and let client poll editor_state.
-            bool actuallyWaited = waitForReady && !compileRequested;
-            if (actuallyWaited)
+            // Earlier Unity versions retain the original behavior.
+#if UNITY_6000_0_OR_NEWER
+            bool shouldWaitForReady = waitForReady && !compileRequested;
+#else
+            bool shouldWaitForReady = waitForReady;
+#endif
+            if (shouldWaitForReady)
             {
                 try
                 {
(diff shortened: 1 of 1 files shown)


dsarno:
@BlindsidedGames Thanks for the contribution and digging into the issue.  I tested it and works fine for me.  @Scriptwonder let us know if this fixes it for you too.  (I reverted the earlier PR changes that weren't actually doing much).

Issue #577, opened 2026-01-18
High performance impact even when MCP server is off


johnkslg:
<img width="1906" height="531" alt="Image" src="https://github.com/user-attachments/assets/38ae6338-9fbb-498b-ac68-33906c65f632" />
I'm seeing spikes of 28ms every one second, it looks like a ton of allocations happening. Makes my game hitch.  
Should at least turn off when mcp is off, as is I have to uninstall the package to remove that hitch.

thomaswilliams02948:
Im having the exact same issue, the game will stutter every second, even during editing mode, while dragging windows around or moving objects in the scene. Removing the mcp package eliminates the hiccups.

Scriptwonder:
@msanatan @dsarno definitely something need to look at.

dsarno:
Yes reviewing the PR and will resolve tomorrow.

dsarno:
@thomaswilliams02948 @johnkslg Can you try again, making sure you are using latest MCP for Unity version (`v9.0.8`)?  I put in a fix for this hitch like a week ago (https://github.com/CoplayDev/unity-mcp/pull/556), and have not been able to reproduce your profiler results in latest releases (at least on Mac -- haven't done full testing on Windows).

dsarno:
Ok this is actually a separate hitch from the UI-related one I fixed, and the one addressed in https://github.com/CoplayDev/unity-mcp/pull/579.  This has to do with the frequency that the Editor State cache is grabbed.  I have a fix for this, may have to wait to tomorrow.  Thanks for flagging this all.

dsarno:
Should be fixed in latest version.  Please let me know if you continue to see the issue. Thanks for flagging.

PR #579, opened 2026-01-19
Eliminate UI stuttering caused by main thread blocking network calls

slepp:
Potentially Fixes GitHub issue #577: 28ms UI stutters every ~1 second when MCP plugin is running

## Root Cause

The `MCPForUnityEditorWindow.OnEditorUpdate()` hook was being called every frame (60+ times/sec) to poll connection status. This triggered `UpdateConnectionStatus()` → `IsLocalHttpServerReachable()`, which makes network socket connection attempts with 50ms timeouts on the main thread.

## Architectural Changes

- Remove `OnEditorUpdate` hook that polled every frame
- Replace continuous polling with event-driven connection status updates:
  - User actions (manual refreshes via `RefreshAllData`)
  - Bridge state changes (via `RequestHealthVerification`)
  - Window focus events (via `OnFocus`)

Connection status now only updates when actually needed.

## Performance Improvements

1. **Eliminated main thread blocking network I/O**
   - Network socket checks no longer execute 60+ times per second
   - Main thread stays responsive for gameplay

2. **TransportCommandDispatcher: Early exit optimization**
   - Early `Pending.Count` check prevents List allocations when idle
   - Before: ~90 allocations/sec → 28ms GC spike every ~1 second
   - After: 0 allocations/sec when idle → no GC stuttering

3. **EditorStateCache: Sequence-based caching**
   - Only increments sequence number when state actually changes
   - Prevents unnecessary `DeepClone()` calls on every `GetSnapshot()`

4. **StdioBridgeHost: Consistent early exit pattern**
   - Added same early `commandQueue.Count` check as TransportCommandDispatcher

## Testing

- Verified MCP plugin runs without stuttering
- Scene hierarchy queries work correctly
- Editor menu items function properly
- Smooth gameplay without frame drops

## Summary by Sourcery

Reduce editor stuttering and GC pressure by avoiding unnecessary per-frame work and allocations in MCP editor integration.

Bug Fixes:
- Stop per-frame connection status polling in the MCP editor window to prevent main-thread blocking network calls that caused UI stutters.
- Avoid repeated DeepClone allocations in editor state snapshots by only updating the cached clone when state actually changes.

Enhancements:
- Short-circuit idle command processing in TransportCommandDispatcher and StdioBridgeHost to eliminate per-frame list allocations when no work is queued.
- Track editor state changes with a sequence-based mechanism so snapshots are only marked updated when relevant conditions change.

Tests:
- Add performance and behavior tests for TransportCommandDispatcher to verify idle operation does not allocate excessively and that commands are still processed correctly under various conditions.

## Summary by CodeRabbit

* **Performance**
  * Reduced garbage collection overhead through smart caching and optimized allocation patterns.
  * Eliminated per-frame allocations and processing when no pending tasks exist.
  * Conditional hook management reduces continuous polling overhead.

* **Improvements**
  * Status monitoring now event-driven rather than continuously polled.

* **Tests**
  * Added comprehensive test suites validating optimization behavior, performance metrics, and error handling scenarios.

### DIFF
--- a/MCPForUnity/Editor/Services/EditorStateCache.cs
+++ b/MCPForUnity/Editor/Services/EditorStateCache.cs
@@ -32,6 +32,15 @@ namespace MCPForUnity.Editor.Services
         private const double MinUpdateIntervalSeconds = 0.25;
 
         private static JObject _cached;
+        private static JObject _cachedClone;
+        private static long _cachedSequence;
+
+        // State tracking to detect when snapshot actually changes
+        private static string _lastTrackedScenePath;
+        private static bool? _lastTrackedIsFocused;
+        private static bool? _lastTrackedIsPlaying;
+        private static bool? _lastTrackedIsUpdating;
+        private static string _lastTrackedActivityPhase;
 
         private sealed class EditorStateSnapshot
         {
@@ -286,7 +295,6 @@ namespace MCPForUnity.Editor.Services
 
         private static JObject BuildSnapshot(string reason)
         {
-            _sequence++;
             _observedUnixMs = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
 
             bool isCompiling = EditorApplication.isCompiling;
@@ -308,6 +316,8 @@ namespace MCPForUnity.Editor.Services
             var testsMode = TestRunStatus.Mode?.ToString();
             string currentJobId = TestJobManager.CurrentJobId;
             bool isFocused = InternalEditorUtility.isApplicationActive;
+            bool isPlaying = EditorApplication.isPlaying;
+            bool isUpdating = EditorApplication.isUpdating;
 
             var activityPhase = "idle";
             if (testsRunning)
@@ -322,7 +332,7 @@ namespace MCPForUnity.Editor.Services
             {
                 activityPhase = "domain_reload";
             }
-            else if (EditorApplication.isUpdating)
+            else if (isUpdating)
             {
                 activityPhase = "asset_import";
             }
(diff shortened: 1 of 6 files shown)


dsarno:
Thanks much. I'm reviewing this, hopefully can test tomorrow and merge.

dsarno:
@slepp Thanks a lot for this PR and the thorough analysis of the performance issues. We were working on the same problem in parallel (it connects to a bunch of other stuff in the guts) and I just merged #595 which addresses #577 with a similar but slightly different approach.

  What we adopted (same core fixes):
  - ✅ Early exit in TransportCommandDispatcher.ProcessQueue() when Pending.Count == 0
  - ✅ Early exit in StdioBridgeHost.ProcessCommands() when commandQueue.Count == 0
  - ✅ State-change tracking in EditorStateCache to only increment sequence when state actually changes
  - ✅ Throttled OnEditorUpdate in the editor window (2-second intervals instead of every frame)

  Where we diverged:

  1. EditorStateCache _cachedClone: We chose to always return a fresh DeepClone() rather than caching it. During code review, we identified a mutation risk—if any caller modifies the returned JObject, it would silently corrupt the cached copy for all subsequent callers. The state-change detection already prevents unnecessary BuildSnapshot() calls, which was the main cost.
  2. TransportCommandDispatcher hook behavior: We intentionally keep the EditorApplication.update hook permanently installed rather than unhooking when idle. This avoids edge cases where Unity must be focused to re-establish contact with background websocket tasks. The early-exit optimization already eliminates the per-frame
  allocations when idle.
  3. MCPForUnityEditorWindow: We throttled OnEditorUpdate to 2-second intervals rather than removing it entirely, as a more conservative approach that still allows periodic status updates.

Thanks again for the contribution -- we'd be glad to include your work in the future, this was just an issue that turned into a rabbit hole for me, so I threw in a bunch of other fixes to my PR too. 🙏

slepp:
Thanks @dsarno for the detailed response. I saw as I dug deeper it was a rabbit hole for sure. Glad you got the chance to bucket it together! And thanks much for this tool, it changed my game development a ton to have it available.

PR #584, opened 2026-01-19
Add automatic uv installation and improve dependency setup flow

hvesuk:
# Description

This PR introduces automatic installation for the `uv` package manager and enhances the overall dependency management experience. Previously, users had to manually install `uv` via external terminals, which often led to "command not found" errors during the first run.

### 🛠 Automatic Installation
Added `InstallUv()` support across all platforms:
*   **Windows**: Automated installation via PowerShell
*   **macOS/Linux**: Automated installation via `curl`

### 🖥 Setup Window Improvements
*   Added an **"Install UV Automatically"** button to the `MCP Setup` window
*   Updated `MCPSetupWindow.uss` to ensure consistent button sizing and better layout alignment (vertical column layout for install links)
*   Buttons remain visible during the setup phase to provide users with all necessary installation options

### 🚀 Smart Startup Flow
*   `ServerManagementService` now proactively checks for `uv` before launching the local server
*   If missing, it prompts the user with a dialog to install it automatically without leaving Unity

### 🔍 Path Resolution Fix
*   Fixed a bug in `PathResolverService` where it would return `"uvx"` instead of `null` when the binary was not found
*   This change ensures the system correctly detects the "Not Found" state, triggering the automated installation flow instead of failing with a shell error

## Benefits
*   **Zero-Config Experience**: Reduces friction for new users by handling dependencies internally
*   **Error Prevention**: Fixes the common `'uvx' is not recognized` shell error
*   **Unified UI**: Provides a more polished and professional setup experience within the Unity Editor

## Summary by Sourcery

Add automatic uv installation support and integrate dependency checks into the local server startup and setup UI to streamline MCP configuration.

New Features:
- Add automatic uv installation on Windows via PowerShell and on macOS/Linux via curl through platform-specific detectors.
- Add an "Install uv automatically" action in the MCP Setup window to trigger the installer from within Unity.
- Prompt users to install uv automatically when starting the local HTTP server if uv is not detected.

Bug Fixes:
- Fix PathResolverService to return null instead of a fallback command when the uvx binary is not found, allowing proper handling of missing uv.

Enhancements:
- Extend the platform detector abstraction to support programmatic uv installation and expose an installation URL for error messaging.
- Improve MCP setup UI layout and button set to keep installation options visible and provide clearer feedback on uv installation success or incomplete detection.

## Summary by CodeRabbit

* **New Features**
  * One-click "Install UV Automatically" button added to setup UI.
  * Automatic UV package manager installation supported on Linux, macOS, and Windows.
  * Server start now prompts and can attempt automatic UV installation if UV is missing.

* **Improvements**
  * Setup window install links reorganized for clearer vertical layout and full-width buttons.


hvesuk:
Hi @dsarno! I’ve opened this PR — let me know if anything needs adjustment. Thanks!

dsarno:
> Hi @dsarno! I’ve opened this PR — let me know if anything needs adjustment. Thanks!

Awesome, thanks for these PRs @hvesuk. I definitely like user friendliness of this flow. Let me play with it when I'm back tomorrow. Also @msanatan  might want to weigh in because he did all the setup window flow.

msanatan:
Hey @hvesuk, thanks for this PR! Previously the setup window had some sort of guidance but it wasn't well structured, so we removed it.

First of all, this works, so well done!

I do have some feedback regarding the UX:

- I don't think we should be running scripts to install software without the user explicitly consenting to it and understand what's happening. Maybe they should see the script being run as well
- This blocks Unity. If `uv` is being installed, it would be nice for users to have an option to cancel, and ideally do other things
- On Windows at least, `uv` doesn't work for me until I restart. I believe that's a common issue for most users. So I'd be more firm in the `Installation Incomplete` dialog with users, make them restart their laptop (explaining why, as you do now)
- It auto-installs on local HTTP, does it do it for stdio? I figure those 2 cases should behave the same
  - That said, I would NOT auto-run this at server startup. Explicit over implicit, even when trying to be convenient. If they try to start the server and `uv` is not detected, they should be prompted to install in a non-blocking way.
- I'm adding a docker command for people to run the server as well. Docker is currently an option to run the server, so maybe enforcing `uv` isn't always appropriate. That said, I have no idea how many users will be inclined to use Docker over `uv`

dsarno:
Hi, are we going to try to land this? If not I'll close in next couple of days. Cc @hvesuk

dsarno:
Closing due to age and inactivity.

PR #639, opened 2026-01-27
Support testpypi

msanatan:
## Type of Change

Save your change type
- New feature (non-breaking change that adds functionality)
- Breaking change (fix or feature that would cause existing functionality to change)
- Documentation update
- Refactoring (no functional changes)
- Test update

## Documentation Updates

## Summary by Sourcery

Add configurable support for using TestPyPI as the package source for the MCP server and persist this preference in editor settings.

New Features:
- Introduce an editor toggle to enable using TestPyPI instead of PyPI when fetching the MCP server package.
- Add a persistent editor preference key to store the TestPyPI usage setting and apply it when constructing uvx arguments.

Enhancements:
- Update advanced settings UI to expose and document the TestPyPI option, including tooltips for clarity.

## Summary by CodeRabbit

* **New Features**
  * Added TestPyPI support: A new toggle in the Advanced settings section allows users to switch to the test Python package repository. When enabled, the system sources packages from TestPyPI, which contains pre-release and experimental versions, providing flexibility for testing and development workflows.

### DIFF
--- a/MCPForUnity/Editor/Clients/Configurators/OpenCodeConfigurator.cs.meta
+++ b/MCPForUnity/Editor/Clients/Configurators/OpenCodeConfigurator.cs.meta
@@ -1,2 +1,11 @@
 fileFormatVersion: 2
 guid: 489f99ffb7e6743e88e3203552c8b37b
+MonoImporter:
+  externalObjects: {}
+  serializedVersion: 2
+  defaultReferences: []
+  executionOrder: 0
+  icon: {instanceID: 0}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 


dsarno:
Thanks for adding this @msanatan! One concern: the Python server and C# client are tightly coupled, so if schemas diverge -- which they do anytime there's a significant server change -- things break silently.    That's the state we're in right now for `main` vs `beta` -- `beta` users are getting fails because they're pulling down the (older) `main` server by default.                                                                                         

Since `beta` auto-publishes to TestPyPI and `main` publishes to PyPI, beta users need TestPyPI to get matching schemas. Making it opt-in means beta is kinda broken by default unless you know to flip this hidden toggle.                                                     
                                                                                                                                      
So we should flip the default to true on the beta branch:
                                                                                                                                                                                                  
> bool useTestPyPI = EditorPrefs.GetBool(EditorPrefKeys.UseTestPyPI, true);                                                           
                                                                                                                                      
That way beta "just works" out of the box, and we flip it back to false when releasing to main.

msanatan:
@dsarno shouldn't it be the other way around? We default for release settings which most customers use, and do patches for beta testers?

dsarno:
I thought that's what I'm suggesting? Default for most users is main with pypi -- or if you opt in to beta via the beta git link download, you get testpypi automatically (and not via a second opt in in the advanced settings, which this PR proposes). Maybe I'm not understanding something about the flow here?

dsarno:
> I thought that's what I'm suggesting? Default for most users is main with pypi -- or if you opt in to beta via the beta git link download, you get testpypi automatically (and not via a second opt in in the advanced settings, which this PR proposes). Maybe I'm not understanding something about the flow here?

@msanatan are we intending to have the main download link default to beta? I think that's what we're doing. But that seems to defeat the point of users opting in to beta and makes everyone a beta tester.  Either way, beta needs to default to testpypi for server or there will be a breaking mismatch between c# and server. 

https://github.com/CoplayDev/unity-mcp.git?path=/MCPForUnity  <-- points to beta.

dsarno:
@msanatan for a more concrete example: 

I just tested the "way it is right now" : `beta` with `PyPI` server.

Because we updated how manage_prefabs work a few days ago (on beta) and the schema changed, manage prefabs is currently broken on beta. The PyPi (older) version bundled with beta uses a schema that doesn't match with the C#:
                                                                                                                                      
  - manage_prefabs(action="open_stage") [old schema] → passes Python validation, but C# rejects: "Unknown action"                                  
  - manage_prefabs(action="modify_contents") [new schema] → fails Python Pydantic validation, never reaches C#                                     
                                                                                                                                      
The schemas are already out of sync. Beta C# expects modify_contents, PyPI 9.2.0 still has open_stage in its enum. Both directions  are broken.                                                                                                                         
                                                                                                                                      
This is why TestPyPI needs to be the default on beta, not opt-in.

dsarno:
Superseded by #640

PR #687, opened 2026-02-05
Fix:ipv6 dualstack listen

whatevertogo:
## Description
Fix IPv4/IPv6 dual-stack handling on Windows. When `localhost` is used as the default host, Windows resolves IPv6 (`::1`) first, which can add 1–2s of delay on some networks. This PR introduces a platform-aware default host policy (Windows uses `127.0.0.1`, other platforms use `localhost`) and centralizes host normalization in a new `HostAddress` helper.

## Type of Change
Bug fix (non-breaking change that fixes an issue)

## Changes Made
- **Add `HostAddress.cs`** – centralized host helper providing:
  - Platform-aware default host (Windows: `127.0.0.1`, others: `localhost`)
  - `NormalizeForClient()` – converts bind-only addresses (`0.0.0.0`, `::`) to a client-usable default
  - `BuildConnectionList()` – ordered connection attempts (IPv4 first)
  - `IsExplicitIPv4/IPv6()` – explicit IP detection
  - Trims surrounding whitespace for more robust config handling
- **Update `ServerManagementService.cs`** – uses `HostAddress.BuildConnectionList()` instead of manual lists
- **Update `WebSocketTransportClient.cs`** – uses `HostAddress.NormalizeForClient()` and `IsBindOnlyAddress()`
- **Update `HttpEndpointUtility.cs`** – uses the platform-aware default local URL
- **Update `Server/src/core/config.py`** – Python server uses the same platform-aware default host policy
- **Add `HostAddressTests.cs`** – 25 unit tests (278 lines) covering all `HostAddress` methods

## Testing/Screenshots/Recordings
- Added `HostAddressTests.cs` with 25 test cases covering:
  - IPv4/IPv6 explicit address detection
  - Bind-only address detection
  - Address normalization
  - Connection list construction
  - Whitespace trimming behavior
  - Platform default host validation

## Documentation Updates
- [ ] I have added/removed/modified tools or resources
- [ ] If yes, I have updated all documentation files using:
  - [ ] The LLM prompt at `tools/UPDATE_DOCS_PROMPT.md` (recommended)
  - [ ] Manual updates following the guide at `tools/UPDATE_DOCS.md`

> Note: This change is primarily internal (helper + wiring); no public MCP tool or resource API changes, so docs updates are optional.

## Related Issues
https://github.com/CoplayDev/unity-mcp/issues/643  
https://github.com/CoplayDev/unity-mcp/issues/672

## Additional Notes
[...]

Issue #672, opened 2026-02-03
Request: developer experience - connect/reconnect automatically, keep server running


bgr:
I'm running MCP for Unity v9.2.0 and `mcpforunityserver@latest` (which based on PyPI should be 9.4.0b20260203071644) so it might just be the wrong combination of versions, but during the last few days of trying to use this I'm experiencing unneeded friction that seems like a quick fix:

* server closes itself when last connection is dropped - fix: just keep the server running
* Unity MCP doesn't connect (start session) automatically - fix: just start a session when there's a server available (this should be easy because it already knows whether to display the "Start Session" button based on server availability)

Note that I'm running the server manually in order to bind on 0.0.0.0 (`uvx.exe --from mcpforunityserver@latest mcp-for-unity --transport http --http-url http://0.0.0.0:4080`) so I would expect it to just stay running forever, but I have to restart it multiple times per hour sometimes, it just exits by itself.

The core of the issue is obviously the constant disconnecting that happens, but I still think the above two things should be the norm and I'm not just suggesting them to work around my current issue. Auto-connect (and reconnect) could be a setting in Unity MCP's UI.

dsarno:
Hi @bgr , we've done some work over the last few days to improve the "sync" between main and beta.  If you're using the main branch, which I recommend, you should be getting the stable PyPi server version, currently `9.4.0`.    Make sure you see the 9.4.0 badge like below, and also that you don't have "Use Beta Server" selected in the Advanced tab.

<img width="798" height="117" alt="Image" src="https://github.com/user-attachments/assets/cb898b1e-ae02-477a-b674-2c55c24b8b81" />

<img width="764" height="602" alt="Image" src="https://github.com/user-attachments/assets/3a2fb70e-bfaf-4d62-9b33-a600937c38ef" />

sballew:
I thought it was just me. I haven't been able to effectively use unity-mcp due to constant disconnects. Trying to restart things results in Claude Code / Cursor being unable to find the Unity session. I'm not using the beta server, though I didn't have success either way.

dsarno:
Thanks for your report. We've just created the beta branch to keep main more stable now that we have more users. But we're still working out a few issues. Hopefully can stabilize today. Thanks.

whatevertogo:
@sballew Sorry to bother you, but could you please share the MCP server logs and the Unity logs? if you have time.

dsarno:
[...]

Issue #664, opened 2026-02-01
[Bug] v.9.3.1 + Claude Code Connection Error


sayplswis:
Hello. I was using version 9.3.0 and everything was working fine, but I updated to 9.3.1 this morning and now I can't connect.

Problem description: I launch the HTTP server (configured for the Claude Code client).
The server starts and the session is active. After that, I launch Claude Code and the server crashes. In the plugin, the "Start Server" button becomes available again and it shows "No Session".

However, the server is still running. Claude Code detects the server, but does not detect the Unity Editor.
If I launch Claude Code first, then launch the server, and then click reconnect, the behavior is the same as the problem described above.

Steps to reproduce

<details>
<summary>Step 1: Start Server</summary>
Unity console
`<b><color=#2EA3FF>MCP-FOR-UNITY</color></b>: Started local HTTP server in terminal: C:\Users\Max1m\.local\bin\uvx.exe --prerelease explicit --from "mcpforunityserver>=0.0.0a0" mcp-for-unity --transport http --http-url http://localhost:8080 --project-scoped-tools --pidfile D:\UNITY\WordRace3D\Library\MCPForUnity\RunState\mcp_http_8080.pid --unity-instance-token f1bdbd414d504db8bf3c742327345a37
UnityEngine.Debug:Log (object)
MCPForUnity.Editor.Helpers.McpLog:Info (string,bool) (at ./Library/PackageCache/com.coplaydev.unity-mcp@f1a5edbd8324/Editor/Helpers/McpLog.cs:40)
MCPForUnity.Editor.Services.ServerManagementService:StartLocalHttpServer () (at ./Library/PackageCache/com.coplaydev.unity-mcp@f1a5edbd8324/Editor/Services/ServerManagementService.cs:318)
MCPForUnity.Editor.Windows.Components.Connection.McpConnectionSection/<OnHttpServerToggleClicked>d__53:MoveNext () (at ./Library/PackageCache/com.coplaydev.unity-mcp@f1a5edbd8324/Editor/Windows/Components/Connection/McpConnectionSection.cs:610)
System.Runtime.CompilerServices.AsyncVoidMethodBuilder:Start<MCPForUnity.Editor.Windows.Components.Connection.McpConnectionSection/<OnHttpServerToggleClicked>d__53> (MCPForUnity.Editor.Windows.Components.Connection.McpConnectionSection/<OnHttpServerToggleClicked>d__53&)
[...]

Issue #643, opened 2026-01-28
beta branch: when Claude Code reconnects, Unity session becomes "no session" even though handshake succeeds. This breaks the workflow completely. Logs attached. But Sometimes it works,and the bug only occurs in beta by using http to connect


whatevertogo:
## the python server:

[01/29/26 00:27:10] INFO     Starting MCP server 'mcp-for-unity-server' with transport 'http' on server.py:2618
                             http://localhost:8080/mcp
INFO:     Started server process [47480]
INFO:     Waiting for application startup.
2026-01-29 00:27:10,235 - docket.worker - INFO - Starting worker 'mylove#47480' with the following tasks:
2026-01-29 00:27:10,235 - docket.worker - INFO - * trace(message: str, ...)
2026-01-29 00:27:10,235 - docket.worker - INFO - * fail(message: str, ...)
2026-01-29 00:27:10,235 - docket.worker - INFO - * sleep(seconds: float, ...)
2026-01-29 00:27:10,236 - mcp.server.streamable_http_manager - INFO - StreamableHTTP session manager started
INFO:     Application startup complete.
INFO:     Uvicorn running on http://localhost:8080 (Press CTRL+C to quit)
INFO:     ('::1', 60557) - "WebSocket /hub/plugin" [accepted]
INFO:     connection open
2026-01-29 00:27:20,308 - mcp.server.streamable_http_manager - INFO - Created new transport with session ID: 4cefe6618a6d4f468265163a4df0d52f
INFO:     ::1:60562 - "POST /mcp HTTP/1.1" 200 OK
INFO:     ::1:60562 - "POST /mcp HTTP/1.1" 202 Accepted
INFO:     ::1:60562 - "GET /mcp HTTP/1.1" 200 OK
2026-01-29 00:27:20,319 - asyncio - ERROR - Task exception was never retrieved
future: <Task finished name='Task-38' coro=<IocpProactor.accept.<locals>.accept_coro() done, defined at C:\Users\18794\AppData\Roaming\uv\python\cpython-3.13.11-windows-x86_64-none\Lib\asyncio\windows_events.py:564> exception=OSError(22, 'The specified network name is no longer available.', None, 64, None)>
Traceback (most recent call last):
  File "C:\Users\18794\AppData\Roaming\uv\python\cpython-3.13.11-windows-x86_64-none\Lib\asyncio\windows_events.py", line 567, in accept_coro
    await future
[...]

PR #688, opened 2026-02-05
fix: harden localhost resolution and reload transport resilience on Windows

dsarno:
### Fix 1: IPv4/IPv6 localhost ambiguity (`#672`)
- Replace default `localhost` references with `127.0.0.1` in bind/connect defaults across Python server and Unity plugin.
- On Windows, `localhost` can resolve to `::1` (IPv6) first, causing server/client address-family mismatch and intermittent disconnects.
- Explicit user-configured `localhost` values are still honored.

### Fix 2: Duplicate TCP listeners after domain reload (`#692`)
- On Windows, domain reload could orphan stdio sockets; permissive socket options then allowed duplicate listeners on the same port.
- Incoming connections could round-robin between live and dead listeners, causing handshake hangs and intermittent failures.
- Root-cause hardening in stdio host/reload path:
  - Stop before domain reload destroys managed state.
  - Centralize stdio teardown sequencing in `StdioBridgeReloadHandler` to avoid resume-flag races.
  - Remove duplicate `beforeAssemblyReload` stop hook from `StdioBridgeHost`.
  - Tighten socket option behavior by platform.
  - Dispose listener socket eagerly during stop.
  - Increase stop/rebind tolerance windows for reload timing.

### Fix 3: HTTP/WebSocket reload resilience hardening
- Domain reload HTTP resume is now bounded-retry (0s, 1s, 3s, 5s, 10s, 30s) instead of one-shot start.
- Before assembly reload, HTTP transport uses synchronous force-stop to avoid leaving orphaned WebSocket resources due to un-awaited async shutdown.
- WebSocket connect now includes localhost loopback fallbacks (`localhost`, then `127.0.0.1`, then `::1`) to handle resolver/address-family mismatches during reconnect/resume.
- Reconnect behavior remains finite (no infinite retry loop).
- Added per-attempt diagnostic logs for HTTP post-reload resume.

### Fix 4: Localhost reachability probe parity (`#643` mitigation)
- Local HTTP reachability probing now treats `localhost` as dual-stack and probes both `127.0.0.1` and `::1`.
- This avoids false "server no longer running" decisions when the server and probe resolve localhost to different loopback families.
- Added characterization tests for probe host candidate generation.

## Issue coverage status
- `#672`: addressed by default host/fallback alignment changes.
- `#692`: addressed by stdio reload/listener lifecycle hardening.
- `#643`: partially mitigated (localhost family mismatch and reload timing paths).
- `#664`: addressed by localhost/IP-family and reload hardening paths included here.
- `#691`: partially mitigated; this PR does **not** claim to fully solve WinError64/accept-path connection loss behavior.

A follow-up PR is planned for the remaining general connection-loss path (including WinError64 mitigation with forward-compatible asyncio startup behavior).

### Files changed
- `Server/src/core/config.py` — `unity_host` default
- `Server/src/main.py` — `--http-url` arg default, host fallbacks, help text
- `MCPForUnity/Editor/Helpers/HttpEndpointUtility.cs` — `DefaultLocalBaseUrl`
- `MCPForUnity/Editor/Services/ServerManagementService.cs` — dual-stack localhost probe candidate builder for local reachability checks
- `MCPForUnity/Editor/Services/Transport/Transports/WebSocketTransportClient.cs` — bind-only host mapping + localhost candidate fallbacks + reconnect handling
- `MCPForUnity/Editor/Services/Transport/Transports/StdioBridgeHost.cs` — socket options, stop disposal, retry params, remove duplicate reload stop hook
- `MCPForUnity/Editor/Services/StdioBridgeReloadHandler.cs` — centralized stdio pre-reload stop sequencing and resume flag handling
- `MCPForUnity/Editor/Services/Transport/TransportManager.cs` — synchronous `ForceStop(TransportMode)` bridge helper
- `MCPForUnity/Editor/Services/HttpBridgeReloadHandler.cs` — synchronous pre-reload stop + bounded post-reload resume retries
- `Server/tests/test_core_infrastructure_characterization.py` — updated assertion + HTTP default/fallback coverage
- `TestProjects/UnityMCPTests/Assets/Tests/EditMode/Services/WebSocketTransportClientTests.cs` — tests for localhost fallback candidate generation
- `TestProjects/UnityMCPTests/Assets/Tests/EditMode/Services/Characterization/ServerManagementServiceCharacterizationTests.cs` — tests for localhost probe host generation

### Relationship to `#687`
PR `#687` by @whatevertogo independently identified the same IPv4/IPv6 root cause. This PR keeps direct default-path changes and also includes reload hardening behavior in the Unity editor-side transport lifecycle.

## Test plan
- [x] All 563 Unity EditMode tests pass (0 failures) on baseline scope before this follow-up hardening
- [x] Python tests pass
- [x] Added/updated Python tests for HTTP default host/URL fallback behavior
- [x] Added Unity EditMode tests for WebSocket localhost fallback candidate generation
- [x] Added Unity characterization tests for localhost dual-stack probe host generation
- [ ] Verify on Windows: single LISTENING socket on port 6400 after multiple domain reloads
- [ ] Verify Unity HTTP session resumes after domain reload with bounded retries (no manual reconnect)
- [ ] Verify explicit `localhost` configs still work (including no false "server no longer running" UI state)
- [ ] Verify bind-only fallback behavior: `0.0.0.0 -> 127.0.0.1`, `:: -> ::1`

Addresses #672 and #692  
Mitigates #643, #664, and #691

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection reliability with IPv4/IPv6 fallbacks for localhost and clearer failure logs.
  * Fixed socket/resource cleanup and added a synchronous teardown path to avoid lingering connections during reloads.
  * Increased retry resilience and wait times for transport startup/shutdown sequences.

* **Other**
  * Default local host addresses and examples now use 127.0.0.1 instead of localhost.

* **Tests**
  * Added unit tests covering connection candidate selection and local probe host behavior.


whatevertogo:
I'd like to suggest https://github.com/CoplayDev/unity-mcp/pull/687 as an alternative approach.

  Main difference: instead of changing defaults everywhere, I added a HostAddress helper that centralizes the logic    
  currently scattered across ServerManagementService, WebSocketTransportClient, and HttpEndpointUtility. It also       
  provides an IPv6 fallback toggle for IPv6-only networks, and aligns Python/C# behavior (domain symmetry pattern).    

  Why #687:
  1. Easier maintenance - future host-related changes only touch one place
  2. IPv6 fallback toggle - helps users on IPv6-only networks without hurting Windows users
  3. Domain symmetry - Python (_get_default_unity_host()) and C# (HostAddress.DefaultHost) share the same
  platform-aware policy
  4. 30 comprehensive tests

  Re: main.py HTTP defaults — Unity explicitly passes --http-url when launching the server (via platform-aware
  HttpEndpointUtility.GetLocalBaseUrl()), so the argparse defaults are only used for manual terminal launches. I kept  
  them as localhost since that's more conventional for standalone CLI tools.

Ultimately, you know the project best — I trust your judgment on which approach fits better here.

dsarno:
I'm putting this PR "on ice" for now while I do more investigation into the root cause.  The elements this touches are core to the mcp package, so I don't want to push "shotgun" fixes to this issue until we understand exactly what's causing it.

parantail:
I don't fully understand the specific details of the fix, but dsarno's branch resolved my issue. Previously, after connecting a session in the Unity Editor, reconnecting from Claude Code would cause the session to disconnect — this fix has resolved that problem.

dsarno:
@parantail Thanks for dropping a note.  When you say my branch, do you mean the **dsarno:fix/localhost-ipv6-resolution** branch that this PR is based on?  That'd be good to know that you've tested it and it's working for you.

parantail:
@dsarno Yes it's **dsarno:fix/localhost-ipv6-resolution**.

Finally, I'm able to open two Unity editors and connect Claude Code sessions to each of them. 😄

dsarno:
@parantail Awesome thanks for the feedback, glad it's working!

PR #690, opened 2026-02-06
fix(run_tests): support snake_case params and handle double-serialized arrays

### LINKED ISSUE #689: run_tests test filtering broken: snake_case params ignored + double-serialized arrays
## Bug Description

The `run_tests` tool's test filtering (`test_names`, `group_names`, `category_names`, `assembly_names`) silently fails — all tests run instead of the filtered subset. This affects any MCP client sending parameters in snake_case (which matches the tool's schema).

### 1. Snake_case parameter names not recognized

The MCP tool schema defines parameters in snake_case (`test_names`, `group_names`, etc.), but `GetFilterOptions()` in `RunTests.cs` only reads camelCase keys:

```csharp
// Line ~149: only checks camelCase
var testNames = ParseStringArray("testNames");
```

Since MCP clients send `test_names` (matching the schema), the lookup returns null and no filter is applied.

**Fix:** Add snake_case fallback to each `ParseStringArray` call:
```csharp
var testNames = ParseStringArray("testNames", "test_names");
```

### 2. Double-serialized arrays not handled

The MCP bridge serializes array parameters as a stringified JSON array inside an outer array. For example, `test_names: ["MyTest"]` arrives as:

```json
{"test_names": ["[\"MyTest\"]"]}
```

The `ParseStringArray` method's Array branch doesn't detect this pattern, so it tries to extract `"[\"MyTest\"]"` as a test name string, which never matches any test.

**Fix:** Add detection for single-element arrays where the element is a stringified JSON array:
```csharp
if (array.Count == 1 && array[0].Type == JTokenType.String)
{
    var inner = array[0].ToString().Trim();
    if (inner.StartsWith("[") && inner.EndsWith("]"))
    {
        try { array = JArray.Parse(inner); }
        catch { /* use original array */ }
    }
}
```

### 3. Same snake_case issue in GetTestJob.cs

`GetTestJob.HandleCommand()` reads `includeDetails` and `includeFailedTests` but the schema uses `include_details` and `include_failed_tests`.

## Affected Files

- `Editor/Tools/RunTests.cs` — `GetFilterOptions()` and `HandleCommand()`
- `Editor/Tools/GetTestJob.cs` — `HandleCommand()`

## Environment

- Unity 6.2 (6000.3)
- unity-mcp v9.4.0
- Claude Code MCP client

bruno1308:
## Summary
- **snake_case parameter support**: `GetFilterOptions()` in `RunTests.cs` only checked camelCase keys (`testNames`, `groupNames`, etc.) but the MCP tool schema defines them in snake_case (`test_names`, `group_names`, etc.). Added fallback lookups so both naming conventions work.
- **Double-serialized array handling**: Some MCP clients (e.g. Claude Code) serialize array parameters as a stringified JSON array inside an outer array (`["[\"name1\"]"]`). Added detection and unwrapping for this pattern in `ParseStringArray()`.
- **Same snake_case fix in `GetTestJob.cs`**: `include_details` and `include_failed_tests` parameters also only checked camelCase.

## Root Cause
When an MCP client sends `test_names: ["MyTest"]`, the C# handler looked up `@params["testNames"]` which returned null, so no filter was applied. The double-serialization issue compounded this — even if the key matched, the array value `["[\"MyTest\"]"]` would extract the stringified JSON as a literal test name, which never matched any test.

## Testing
Verified with Claude Code MCP client:
- Before: `run_tests(test_names=["SomeTest"])` always ran 0 tests
- After: correctly filters to only the specified tests

Fixes #689

## Summary by Sourcery

Improve MCP test commands parameter handling for better compatibility with different clients.

Bug Fixes:
- Allow run_tests and get_test_job commands to read both camelCase and snake_case parameter names for filter and include options.
- Correctly parse array parameters that may be provided as stringified JSON arrays, double-serialized arrays, or nested arrays so test filters apply as intended.

## Summary by CodeRabbit

* **Refactor**
  * Consolidated parameter handling for test operations to improve code consistency and reduce duplication across testing utilities.

### DIFF
--- a/MCPForUnity/Editor/Helpers/ParamCoercion.cs
+++ b/MCPForUnity/Editor/Helpers/ParamCoercion.cs
@@ -1,7 +1,5 @@
 using System;
 using System.Globalization;
-using System.Linq;
-using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
 namespace MCPForUnity.Editor.Helpers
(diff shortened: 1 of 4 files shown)


bgr:
I've applied the changed files in this PR (as of 854b42195a) straight into my `Library\PackageCache\com.coplaydev.unity-mcp@418cf0dcc873\Editor` (v9.4.1-beta.1) and can confirm that Claude is happily using test filtering now.

msanatan:
Hi @bruno1308, thanks for this! Can you look at `ToolParams.cs` - instead of creating a new GetParam function you can probably reuse GetRaw or otherwise. Happy to merge otherwise

You can probably move `CoerceStringArray` to `ToolParams.cs` and keep everything central tbf

bruno1308:
Hey @msanatan, thanks for the feedback! I've pushed the requested changes in fa2cd19:

1. **Removed `GetParam` from `ParamCoercion`** — `ToolParams.GetRaw()` already handles snake_case/camelCase fallback via `StringCaseUtility`, so it was redundant.

2. **Moved `CoerceStringArray` into `ToolParams.cs`** — added as an `internal static` helper with a new public `GetStringArray(key)` method that chains through `GetToken()` for automatic case fallback.

3. **Updated `RunTests.cs` and `GetTestJob.cs`** to use `new ToolParams(@params)` with `.GetBool()` and `.GetStringArray()` instead of the manual `ParamCoercion.GetParam()` + `ParamCoercion.CoerceStringArray()` calls.

No functionality was lost — the coercion logic is identical, and null params are already normalized to `new JObject()` by `CommandRegistry` and `TransportCommandDispatcher` upstream.

msanatan:
Thanks @bruno1308 . If you want a shout on LinkedIn/Twitter/Discord on the release then please share your username!

PR #798, opened 2026-02-20
HTTP transport: respect retry_on_reload and add stale connection detection

### LINKED ISSUE #795: HTTP transport ignores retry_on_reload and lacks stale connection detection
## Summary

The HTTP/WebSocket transport path (`PluginHub`) does not respect `retry_on_reload=False` and lacks the stale connection detection that the stdio path now has. This means script mutations sent via HTTP clients (Claude Code, Cursor, etc.) can be retried during domain reload, causing duplicate execution — the same class of bug fixed for stdio in PR #792.

## Root Cause

`send_with_unity_instance()` discards all kwargs (including `retry_on_reload`) when routing through the HTTP path. It extracts `command_type` and `params`, then calls `PluginHub.send_command_for_instance()` which has no knowledge of retry semantics.

```python
# unity_transport.py — HTTP branch
command_type = args[0]
params = args[1] if len(args) > 1 else kwargs.get("params")
# ... retry_on_reload is silently dropped
raw = await PluginHub.send_command_for_instance(unity_instance, command_type, params)
```

## Specific Issues

| Issue | Stdio (fixed) | HTTP (broken) |
|-------|--------------|---------------|
| Stale connection detection before send | `_ensure_live_connection()` MSG_PEEK | None — detected only via 30s timeout |
| `retry_on_reload=False` respected | Yes | **No — silently dropped** |
| Reconnect wait during mutations | Skipped when `retry_on_reload=False` | Always waits 20s for session reconnect |
| Duplicate mutation risk | Low | **High** |
| Connection-lost verification | SHA comparison after reconnect | Timeout → client may retry blindly |

## Suggested Fix

1. **Pass `retry_on_reload` through the HTTP path** — `PluginHub.send_command_for_instance()` should accept and respect it. When False, skip `_resolve_session_id()` wait and don't retry on disconnect.

2. **Add pre-send connection validation** — equivalent to stdio's `_ensure_live_connection()`, probe WebSocket liveness before sending mutations.

3. **Unify the transport retry contract** — Both stdio and HTTP paths implement their own retry/reconnect/reload-wait logic independently, which means every fix has to be applied twice and debugged twice. Consider extracting a shared `TransportPolicy` or similar abstraction that both paths use:
   - Stale connection detection (pre-send health check)
   - Reload-aware retry semantics (`retry_on_reload` flag)  
   - Post-mutation wait-for-ready
   - Connection-lost-after-send detection
   
   This doesn't need to be a deep abstraction — even a shared set of helper functions that both paths call at the right points would prevent the current situation where stdio gets a fix and HTTP silently remains broken.

## Context

Discovered during PR #792 (fix script edit tools retrying non-idempotent commands during reload). The stdio path was fixed end-to-end; this issue tracks the equivalent work for HTTP.

dsarno:
## Summary

Fixes #795 — the HTTP/WebSocket transport path (PluginHub) now respects retry_on_reload=False and performs pre-send liveness checks, matching the stdio path behavior from PR #792.

- **unity_transport.py**: Extract retry_on_reload from kwargs and forward to PluginHub.send_command_for_instance() (previously silently dropped)
- **plugin_hub.py**:
  - _resolve_session_id() accepts retry_on_reload; when False, skips the reconnect wait loop
  - send_command_for_instance() accepts retry_on_reload; when False + stale connection, returns failure immediately instead of retrying
  - New _ensure_live_connection() checks WebSocket state before sending (best-effort pre-send liveness, mirrors stdios MSG_PEEK check)
  - New _evict_connection() proactively cleans up stale sessions
  - Skips fast-fail readiness probe when retry_on_reload=False

## Test plan

- [x] 3 new unit/integration tests (707/707 passing)
- [x] Smoke tested against live Unity instance via HTTP transport:
  - script_apply_edits (uses retry_on_reload=False) — survives domain reload
  - create_script (uses retry_on_reload=False) — survives domain reload
  - refresh_unity with compile — recovers from WebSocket disconnect
  - No compilation errors after mutations

Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Codex <noreply@openai.com>

## Summary by Sourcery

Respect retry_on_reload for HTTP/WebSocket Unity transport and add stale connection detection and eviction to improve domain reload behavior and error signalling.

Bug Fixes:
- Ensure HTTP transport forwards retry_on_reload to PluginHub so commands can opt out of reconnect waits during Unity domain reloads.
- Avoid sending commands over stale WebSocket sessions by checking liveness before send and failing fast when retry_on_reload is disabled.

Enhancements:
- Introduce explicit stale connection eviction in PluginHub, cleaning up WebSocket state, background ping tasks, and pending futures when a session is no longer valid.
- Add a shared unavailable retry response helper to standardize error payloads when Unity sessions are missing or connections are stale.
- Gate fast-fail readiness probes on retry_on_reload to avoid unnecessary waits when callers request no reload retries.

Tests:
- Add integration tests covering no-wait behavior when retry_on_reload is disabled, fast-fail on stale HTTP sessions without retry, and HTTP transport propagation of retry_on_reload.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection stability with proactive liveness checks and automatic cleanup of stale connections.
  * Enhanced retry control during domain reloads to enable faster fallbacks when connections are unavailable.
  * Standardized error responses for unavailable sessions.

* **Tests**
  * Added integration tests for retry behavior control and stale connection detection.
  * Added HTTP transport verification tests.

### DIFF
--- a/Server/src/transport/plugin_hub.py
+++ b/Server/src/transport/plugin_hub.py
@@ -10,7 +10,7 @@ import uuid
 from typing import Any, ClassVar
 
 from starlette.endpoints import WebSocketEndpoint
-from starlette.websockets import WebSocket
+from starlette.websockets import WebSocket, WebSocketState
 
 from core.config import config
 from core.constants import API_KEY_HEADER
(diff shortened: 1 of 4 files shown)


dsarno:
@coderabbitai review

PR #981, opened 2026-03-24
Replace session-global Unity instance selection with explicit routing

CharlieHess:
## Summary
`set_active_instance` works for simple single-editor flows, but it breaks down as soon as multiple Unity editors are connected at the same time. The active instance lives in mutable session state, which means one agent can retarget another agent's calls, and an unbound session can drift onto the wrong editor.

The root problem is that instance selection is modeled as ambient state instead of request data. The server can resolve a target internally, but the MCP surface did not make that routing explicit, and the unbound HTTP endpoint had no safe way to pin a whole session to one editor.

This PR makes instance routing explicit. Unity-backed tools and resources now accept `unity_instance`, HTTP clients can bind an entire session to `/mcp/instance/{Name@hash}`, and `set_active_instance` is demoted to a compatibility fallback instead of the primary multi-editor pattern. The middleware also stops silently selecting a different editor when multiple instances are connected, which makes concurrent-editor workflows predictable and safe.

The test coverage here focuses on the contract reviewers care about: the new tool/resource metadata, bound endpoint behavior, and the failure mode when multiple editors are available but no explicit target is provided.

## Summary by Sourcery

Make Unity instance routing explicit and safe across HTTP and stdio transports, replacing implicit session-global selection with per-request and endpoint-bound targeting.

New Features:
- Expose an instance-bound HTTP MCP endpoint that pins a session to a specific Unity editor via /mcp/instance/{Name@hash}.
- Add explicit unity_instance targeting to Unity-backed tools and resources, including resource URI query support.
- Provide compatibility handling so set_active_instance rejects use on instance-bound endpoints and remains a fallback for simple single-editor workflows.

Enhancements:
- Refine Unity instance middleware to prefer explicit per-call, bound-endpoint, or compatibility defaults over implicit autoselection, and prevent overrides on bound endpoints.
- Update tool and resource registries to inject an optional unity_instance parameter and annotate unity-targeted resources and tools in their metadata.
- Tighten stdio legacy instance resolution to require explicit selection when multiple Unity instances are running, instead of auto-selecting the most recent.
- Improve compatibility with different FastMCP layouts when resolving HTTP request and header dependencies.
- Adjust tool-list filtering to respect bound Unity instances and ignore stale session-selected instances.
- Clarify instance-targeting behavior and failure modes in the server instructions and user-facing error messages.

Tests:
- Add integration and unit tests for inline and bound unity_instance routing, including override rejection on instance-bound endpoints and bound routing behavior.
- Add tests covering the new tool and resource registry metadata for unity_instance parameters and URI templates.
- Add tests for the new instance-bound MCP HTTP route wiring and for legacy stdio instance selection behavior when multiple instances exist.
- Update existing transport characterization tests to reflect user_id-only injection in list_tools and the new filtering behavior.

## Release Notes

* **New Features**
  * Added instance-bound HTTP endpoints at `/mcp/instance/{instance}` for direct targeting of specific Unity instances via URL path.
  * Enabled `unity_instance` parameter for explicit instance selection in function calls as an alternative to `set_active_instance`.

* **Behavior Changes**
  * Instance selection now requires explicit specification when multiple instances are available instead of auto-selecting the most recent.
  * Improved error messages providing clearer guidance on resolving instance ambiguity.

* **Documentation Resources**
  * Marked GameObject, Prefab, Tool Groups, and Unity Instances resources as server-only (no longer instance-specific).

### DIFF
--- a/Server/src/main.py
+++ b/Server/src/main.py
@@ -15,10 +15,11 @@ from services.custom_tool_service import (
     resolve_project_id_for_unity_instance,
 )
 from core.config import config
-from starlette.routing import WebSocketRoute
+from starlette.routing import Route, WebSocketRoute
 from starlette.responses import JSONResponse
 import argparse
 import asyncio
+import uvicorn
 
 # Fix to IPV4 Connection Issue #853
 # Will disable features in ProactorEventLoop including subprocess pipes and named pipes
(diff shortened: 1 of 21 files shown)


jordandunmire97-ai (approved):
This is amazing

Scriptwonder:

PR #1057, opened 2026-04-11
Replace localhost with 127.0.0.1 for correct Codex handshake

hvesuk:
## Description
- update Codex HTTP MCP config generation to use `127.0.0.1` instead of `localhost`

## Why
In practice, Unity MCP was reachable and responded correctly, but Codex could fail during the MCP initialize/handshake step when configured with `http://localhost:8080/mcp`.

Using `http://127.0.0.1:8080/mcp` fixes the issue consistently after restarting Codex, which suggests the problem is client-side loopback/hostname resolution behavior in Codex rather than a Unity MCP server issue.

## Type of Change

Save your change type
- Refactoring (no functional changes)

## Changes Made
- update Codex HTTP MCP config generation to use `127.0.0.1` instead of `localhost`
- keep the change scoped to `CodexConfigHelper` only
- leave the shared HTTP endpoint logic unchanged for other clients

- enable HTTP transport for Codex config generation
- verify generated TOML contains:

  `[mcp_servers.unityMCP]`
  `url = "http://127.0.0.1:8080/mcp"`

- restart Codex
- confirm Codex connects to Unity MCP successfully
- confirm no behavior changes for non-Codex clients

## Documentation Updates

## Summary by Sourcery

Update Codex HTTP MCP configuration generation to use a loopback IP address that ensures reliable Codex handshakes.

Bug Fixes:
- Ensure Codex connects reliably to Unity MCP by avoiding hostname resolution issues when using HTTP transport.

Enhancements:
- Introduce a Codex-specific HTTP URL helper to adjust the generated MCP RPC URL without affecting other clients.

## Summary by CodeRabbit

* **Bug Fixes**
  * Normalize loopback hostnames by replacing "localhost" with "127.0.0.1" when configuring HTTP/SSE endpoints.
  * Apply this normalization consistently across HTTP transport configuration paths so generated URLs use the loopback IP.
  * No behavior changes for stdio transport.

### DIFF
--- a/MCPForUnity/Editor/Helpers/CodexConfigHelper.cs
+++ b/MCPForUnity/Editor/Helpers/CodexConfigHelper.cs
@@ -36,7 +36,7 @@ namespace MCPForUnity.Editor.Helpers
             if (useHttpTransport)
             {
                 // HTTP mode: Use url field
-                string httpUrl = HttpEndpointUtility.GetMcpRpcUrl();
+                string httpUrl = GetCodexHttpUrl();
                 unityMCP["url"] = new TomlString { Value = httpUrl };
 
                 // Enable Codex's Rust MCP client for HTTP/SSE transport
@@ -191,7 +191,7 @@ namespace MCPForUnity.Editor.Helpers
             if (useHttpTransport)
             {
                 // HTTP mode: Use url field
-                string httpUrl = HttpEndpointUtility.GetMcpRpcUrl();
+                string httpUrl = GetCodexHttpUrl();
                 unityMCP["url"] = new TomlString { Value = httpUrl };
             }
             else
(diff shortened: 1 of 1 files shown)


Scriptwonder:
Hi Alex, thanks for the fix on the Codex issue. Can you add comments on this hard-coded fix, in case Codex in the future route it correctly, and this hard-coded fix becomes an issue?

hvesuk:
@Scriptwonder I added a comment. Is that enough?

Scriptwonder:
I looked into it more closely and implemented a more generalized solution in #1084; this way, when Codex and other similar clients that are running IPv6 are running local HTTP, it will transform local to 127.0.0.1, all in one place (tested with Windows Codex App). Thanks for submitting and putting in the good work!

hvesuk:
@Scriptwonder Thanks!

PR #1075, opened 2026-04-25
Fix Unity 6000.6.x object find compatibility

rurush47:
## Summary
- Fix package compilation errors in Unity 6000.6.x
- Adds a UnityFindObjectsCompat helper for Object.FindObjectsByType API changes across Unity 2022 and Unity 6.
- Replaces direct FindObjectsByType(..., FindObjectsSortMode.None) calls to avoid Unity 6.5+ deprecation warnings while preserving older Unity support.

## Testing
- Ran git diff --check.
- Verified the same compatibility approach is already present on the beta branch.


Scriptwonder:
We actually had this fix in the beta branch for a few weeks now but have not patched it to the main. Will run a patch to the main, and close this one. Thanks for the work.

PR #1076, opened 2026-04-25
Fix create_script validator false-positive inside method bodies

justinpbarnett:
## Summary
- Fix `ManageScript` duplicate-method validation so it only counts method declarations that are direct members of a type.
- Prevent repeated method calls such as `return Compute();` from being misclassified as duplicate method declarations.
- Add regression coverage for repeated invocations and same-signature local functions in separate method bodies.

## Test plan
- [x] Reproduced on `upstream/beta` with a temporary Unity `-executeMethod` harness: repeated `return Compute();` calls produced duplicate `Compute` diagnostics.
- [x] Verified the same harness passes on this branch with `REPRO_OK`.
- [x] `git diff --check`

## Notes
- This follows up the same validator area fixed in #1045.
- No tool/resource schema or documentation updates are needed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved duplicate method signature detection to more accurately identify actual duplicate declarations while avoiding false positives from local functions in different scopes and method invocations within method bodies.

* **Tests**
  * Added test coverage for edge cases in duplicate method signature detection.

### DIFF
--- a/MCPForUnity/Editor/Tools/ManageScript.cs
+++ b/MCPForUnity/Editor/Tools/ManageScript.cs
@@ -2714,12 +2714,18 @@ namespace MCPForUnity.Editor.Tools
             }
             // Second pass: build per-position containing type array
             var containingTypeArr = new string[codeOnly.Length];
+            var braceDepthArr = new int[codeOnly.Length];
+            var typeMemberDepthArr = new int[codeOnly.Length];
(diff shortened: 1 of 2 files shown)


Scriptwonder:
Great to see you Justin! Thanks for the fix :)

PR #1089, opened 2026-04-27
fix: support Unity 2022.2 object lookup compatibility

chenkunqing:
## Description

Fix a Unity 2022.2 compatibility issue in `UnityFindObjectsCompat`.

Some Unity 2022.2 editor versions do not reliably expose the newer object lookup APIs used by the package. This caused compilation failures in projects using Unity 2022.2.1f1.

## Type of Change

## Changes Made

- Delayed use of `Object.FindObjectsByType(...)` until `UNITY_2022_3_OR_NEWER`
- Delayed use of `Object.FindAnyObjectByType(...)` until `UNITY_2022_3_OR_NEWER`
- Kept `FindObjectsOfType(...)` / `FindObjectOfType(...)` as the fallback path for Unity 2022.2.x
- Added local pragma guards for obsolete fallback APIs to keep the compatibility wrapper warning-clean

- Reproduced the compile failure on Unity `2022.2.1f1`
- Verified the project compiles cleanly in Unity `2022.2.1f1` after the compatibility change
- Ran `git diff --check`

## Documentation Updates

## Related Issues

Relates to Unity 2022.2.x compilation failures in `UnityFindObjectsCompat`.

## Additional Notes

This change intentionally keeps the fix minimal and only adjusts the version gating for object lookup APIs in the compatibility helper.

## Summary by CodeRabbit

* **Chores**
  * Updated Unity compatibility layer to utilize newer APIs for Unity 2022.3+ while maintaining backward compatibility with older versions.
  * Enhanced code stability with improved warning handling for legacy code paths.

### DIFF
--- a/MCPForUnity/Runtime/Helpers/UnityFindObjectsCompat.cs
+++ b/MCPForUnity/Runtime/Helpers/UnityFindObjectsCompat.cs
@@ -4,11 +4,9 @@ using UObject = UnityEngine.Object;
 namespace MCPForUnity.Runtime.Helpers
 {
     /// <summary>
-    /// Version-compatible wrappers for the Object.FindObjectsOfType / FindObjectsByType family,
-    /// which changed across Unity 2022 → 6.0 → 6.5:
-    ///   Pre-2022.2  : FindObjectsOfType / FindObjectOfType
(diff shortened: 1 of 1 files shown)


Scriptwonder:
This is integrated in latest patch #1097, since we need to think about how other APIs should be taken care of. Thanks for bringing it up!

PR #1116, opened 2026-05-09
fix(roslyn): install missing System.Runtime.CompilerServices.Unsafe v6 + surface inner errors

sMartz1:
## Summary

The `RoslynInstaller` ships only 4 NuGet packages but `Microsoft.CodeAnalysis` 4.12.0 on netstandard2.0 also references `System.Runtime.CompilerServices.Unsafe` **v6.0.0.0**, which is **not** what Unity ships (Unity bundles a v4.x of that assembly). The reference is unresolved at runtime, so:

```
Roslyn.Utilities.StringTable..cctor()
  → SegmentedArray<T>..ctor()
    → FileNotFoundException: System.Runtime.CompilerServices.Unsafe, Version=6.0.0.0
→ TypeInitializationException: 'Roslyn.Utilities.StringTable'
  → TypeInitializationException: 'Microsoft.CodeAnalysis.CSharp.CSharpSyntaxTree'
    → every parse / compile attempt throws TargetInvocationException
```

Worse, the failure is invisible: `RoslynCompiler.Compile`'s catch block logs only `e.Message`, and for `TargetInvocationException` that string is the generic _"Exception has been thrown by the target of an invocation."_ — the real cause in `InnerException` is silently dropped, so the install looks fine and `IsAvailable` returns `true`, but every actual compile fails with a useless message.

## Repro

1. Fresh project, run **Tools → MCP for Unity → Install Roslyn** (or however the install is triggered).
2. Use the MCP `execute_code` tool with `compiler="roslyn"` (or `auto`) on any C# 7+ snippet.
3. Result: `Compilation failed: Roslyn compilation error: Exception has been thrown by the target of an invocation.` — no other detail.
4. Drilling via reflection (replicating `Compile` step-by-step with proper `InnerException` walking) reveals the real chain shown above.

## Fix

* **`MCPForUnity/Editor/Setup/RoslynInstaller.cs`** — add the missing transitive dependency `system.runtime.compilerservices.unsafe` v6.0.0 to `NuGetEntries` so a fresh install drops 5 DLLs into `Assets/Plugins/Roslyn/` instead of 4. This makes `IsInstalled()` accurately reflect "Roslyn will actually load".
* **`MCPForUnity/Editor/Tools/ExecuteCode.cs`** — in `RoslynCompiler.Compile`'s outer catch, walk the `InnerException` chain to the deepest cause and report `Type.Name + Message` of that root, instead of the generic wrapper. This makes future bootstrap regressions debuggable from the tool output alone.

## Verification

Tested on Unity 2022.3 / Windows / netstandard2.0 build target:

1. Before patch: Roslyn `IsAvailable=True`, every parse fails with the generic message.
2. Removed local `Assets/Plugins/Roslyn/`, ran the patched installer → 5 DLLs present.
3. Forced a domain reload (the cached `TypeInitializationException` would otherwise persist for the AppDomain lifetime).
4. After patch: `execute_code` with `compiler="roslyn"` compiles C# 7+ snippets (`var`, `?.`, `??`, string interpolation) and runs them successfully.

## Notes

* The transitive dep is also pulled by `System.Reflection.Metadata` 8.0.0 and `System.Memory`, but the Roslyn IL references the specific `Version=6.0.0.0` ref-assembly — there is no binding redirect, and Unity's bundled `System.Runtime.CompilerServices.Unsafe` doesn't satisfy a strong-named v6 reference.
* No behavior changes for users who already have Roslyn working (e.g. via a binding redirect or a different plugins layout) — the new entry is additive.
* The `catch` change is purely diagnostic; it never alters whether compilation succeeds, only what gets surfaced when it doesn't.

## Test plan

- [ ] On a fresh project (no `Assets/Plugins/Roslyn/`), run the installer → confirm 5 DLLs are present.
- [ ] Restart the Editor (force domain reload).
- [ ] Execute a Roslyn-only snippet via `execute_code` → confirm success.
- [ ] (Negative) Manually delete `System.Runtime.CompilerServices.Unsafe.dll` and try again → confirm the error message now reports `FileNotFoundException` directly instead of the generic invocation wrapper.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compiler error reporting to surface the root exception type and message for clearer troubleshooting.
  * Installer now verifies on-disk assembly versions and treats missing/outdated or unreadable assemblies as not installed, ensuring correct reinstalls.

* **Chores**
  * Added required runtime dependency to the installation list and adjusted related installer configuration.

[Review Change Stack](https://app.coderabbit.ai/change-stack/CoplayDev/unity-mcp/pull/1116)

### DIFF
--- a/MCPForUnity/Editor/Setup/RoslynInstaller.cs
+++ b/MCPForUnity/Editor/Setup/RoslynInstaller.cs
@@ -13,10 +13,14 @@ namespace MCPForUnity.Editor.Setup
 
         private static readonly (string packageId, string version, string dllPath, string dllName)[] NuGetEntries =
         {
-            ("microsoft.codeanalysis.common",    "4.12.0", "lib/netstandard2.0/Microsoft.CodeAnalysis.dll",       "Microsoft.CodeAnalysis.dll"),
-            ("microsoft.codeanalysis.csharp",    "4.12.0", "lib/netstandard2.0/Microsoft.CodeAnalysis.CSharp.dll","Microsoft.CodeAnalysis.CSharp.dll"),
-            ("system.collections.immutable",     "8.0.0",  "lib/netstandard2.0/System.Collections.Immutable.dll", "System.Collections.Immutable.dll"),
-            ("system.reflection.metadata",       "8.0.0",  "lib/netstandard2.0/System.Reflection.Metadata.dll",   "System.Reflection.Metadata.dll"),
+            ("microsoft.codeanalysis.common",         "4.12.0", "lib/netstandard2.0/Microsoft.CodeAnalysis.dll",                   "Microsoft.CodeAnalysis.dll"),
+            ("microsoft.codeanalysis.csharp",         "4.12.0", "lib/netstandard2.0/Microsoft.CodeAnalysis.CSharp.dll",            "Microsoft.CodeAnalysis.CSharp.dll"),
+            ("system.collections.immutable",          "8.0.0",  "lib/netstandard2.0/System.Collections.Immutable.dll",             "System.Collections.Immutable.dll"),
+            ("system.reflection.metadata",            "8.0.0",  "lib/netstandard2.0/System.Reflection.Metadata.dll",               "System.Reflection.Metadata.dll"),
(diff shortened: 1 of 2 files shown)


Scriptwonder:
Thanks for this PR! Much needed since we dont have benchmark tools for roslyn.

PR #1117, opened 2026-05-10
feat: configurable macOS terminal app via EditorPrefs

skyhills13:
## Summary
- The macOS branch of `TerminalLauncher.CreateTerminalProcessStartInfo` hardcodes `open -a Terminal`, so Terminal.app spawns even when the user runs iTerm2 / Warp / Alacritty / etc. macOS LaunchServices resolves `-a Terminal` by app name, so iTerm's *Make iTerm2 default term* setting cannot redirect it — the package code itself has to expose the choice.
- Read the terminal app name from a new EditorPref (`MCPForUnity.MacOSTerminalApp`). Unset/empty/whitespace → falls back to `"Terminal"`, so existing users see no change.
- Quote the app name in the `-a` argument so values with spaces (e.g. `iTerm 2`) work.
- Register the key in `EditorPrefsWindow`'s `knownPrefTypes` (String) so it shows up in the existing EditorPrefs UI — no new window/menu surface needed.

## Motivation (concrete case)
In my own project I had to embed this package and hand-edit the line to `-a iTerm` to stop Terminal.app from popping up on every server restart. That hand-edit is what this PR generalizes: same root change, but driven by `EditorPrefs` so it survives package upgrades and works for any macOS terminal app the user prefers.

## How to use
1. *Window → MCP for Unity → EditorPrefs* (or whichever entry shows the EditorPrefs window).
2. Set `MCPForUnity.MacOSTerminalApp` to the bundle name you want, e.g. `iTerm`, `Warp`, `Alacritty`. Same value you would pass to `open -a <name>`.
3. Restart the MCP server. The new terminal window opens in the chosen app.

## Compatibility
- Default behavior (key unset, empty, or whitespace-only) → `Terminal`. Identical to current behavior.
- macOS only. Windows and Linux branches unchanged.

## Test plan (not yet executed in this PR — happy to do whatever the maintainers want before merge)
- [x] macOS: set `MCPForUnity.MacOSTerminalApp = "iTerm"`, restart server → iTerm window opens with the MCP server log.
- [x] macOS: clear the EditorPref → Terminal.app opens (regression check).
- [x] macOS: name with a space (e.g. `iTerm 2`) — quoting handles it.
- [ ] Editor build check on Windows / Linux (no behavioral change there; just confirm `using UnityEditor;` is fine — it already is in this codebase, so this is a formality).

The only verification I have done so far is the equivalent hand-patched line (`-a iTerm` hardcoded) running in my own project against the current beta package, which confirms `open -a <app>` is the right control point. The EditorPrefs-driven path in this PR has not yet been exercised in a live Unity build.

## Notes for reviewers
- I deliberately avoided adding a new dedicated settings UI — the existing EditorPrefs window already lists registered keys, so adding the key to `knownPrefTypes` is enough surface for now. Happy to wire a dropdown into the main MCP window if you'd prefer something more discoverable.
- No tests added. The macOS branch is `#if UNITY_EDITOR_OSX` and the new logic is `EditorPrefs.GetString` + a fallback string check, which is hard to unit-test without abstracting EditorPrefs. If a small helper extraction is preferred for testability, let me know and I'll add it.

## Summary by CodeRabbit

* **New Features**
  * macOS users can configure their preferred terminal app via a new editor preference. If unset, the editor defaults to the standard Terminal app for compatibility.
* **Tests**
  * Added macOS-focused tests validating terminal selection fallback, trimming behavior, and correct quoting of app names and script paths containing spaces.

### DIFF
--- a/MCPForUnity/Editor/Constants/EditorPrefKeys.cs
+++ b/MCPForUnity/Editor/Constants/EditorPrefKeys.cs
@@ -71,5 +71,7 @@ namespace MCPForUnity.Editor.Constants
         internal const string LogRecordEnabled = "MCPForUnity.LogRecordEnabled";
 
         internal const string ExecuteCodeCompiler = "MCPForUnity.ExecuteCode.Compiler";
+
+        internal const string MacOSTerminalApp = "MCPForUnity.MacOSTerminalApp";
     }
 }
--- a/MCPForUnity/Editor/Services/Server/TerminalLauncher.cs
+++ b/MCPForUnity/Editor/Services/Server/TerminalLauncher.cs
@@ -1,6 +1,8 @@
 using System;
 using System.IO;
+using MCPForUnity.Editor.Constants;
 using MCPForUnity.Editor.Helpers;
+using UnityEditor;
 using UnityEngine;
 
 namespace MCPForUnity.Editor.Services.Server
@@ -45,10 +47,12 @@ namespace MCPForUnity.Editor.Services.Server
                 "clear\n" +
                 $"{command}\n");
             ExecPath.TryRun("/bin/chmod", $"+x \"{scriptPath}\"", Application.dataPath, out _, out _, 3000);
+            string terminalApp = EditorPrefs.GetString(EditorPrefKeys.MacOSTerminalApp, "Terminal");
+            if (string.IsNullOrWhiteSpace(terminalApp)) terminalApp = "Terminal";
             return new System.Diagnostics.ProcessStartInfo
             {
                 FileName = "/usr/bin/open",
-                Arguments = $"-a Terminal \"{scriptPath}\"",
+                Arguments = $"-a \"{terminalApp}\" \"{scriptPath}\"",
                 UseShellExecute = false,
                 CreateNoWindow = true
             };
--- a/MCPForUnity/Editor/Services/Server/TerminalLauncher.cs
+++ b/MCPForUnity/Editor/Services/Server/TerminalLauncher.cs
@@ -27,6 +27,21 @@ namespace MCPForUnity.Editor.Services.Server
             }
         }
 
+        /// <summary>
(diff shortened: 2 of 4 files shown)


Scriptwonder:
Hi thanks for the PR! Late to leave this review but do you think you can fulfill your test plan?  Just want to double check before merging this.

dooby135:
Thanks for the nudge! I've addressed the testability gap from my original notes.

Rather than leaving the macOS logic untestable behind `#if UNITY_EDITOR_OSX`, I extracted the two pure pieces of behavior into `internal static` helpers on `TerminalLauncher`:

- `ResolveMacTerminalApp(configuredApp)` — the EditorPrefs fallback (unset/empty/whitespace → `Terminal`, otherwise trimmed value)
- `BuildMacOpenArguments(configuredApp, scriptPath)` — the `open -a "<app>" "<script>"` argument string

The `#if UNITY_EDITOR_OSX` branch now reads the EditorPref and routes through these, so behavior is unchanged but the decision logic is unit-testable without a live Editor or EditorPrefs.

Added EditMode tests in `TerminalLauncherTests` covering the test-plan cases:
- unset / empty / whitespace-only → falls back to `Terminal` (regression check)
- custom app names (`iTerm`, `Warp`) are used
- quoting handles app names **and** script paths containing spaces (e.g. `iTerm 2`, `/Users/me/My Project/...`)

These are platform-agnostic pure functions, so the tests run on the Linux CI runner too — no macOS-only gating needed. The one thing automated tests can't assert is "an iTerm window visually appears," but that's purely `open -a`'s job; the only thing this PR changes is the argument string, which is now fully covered.

Could you re-apply `safe-to-test` so the Unity Tests workflow runs against the new commit? (Pushed as `edfb2da`.)

+ I didn't have a window machine so I cannot check last test plan

Scriptwonder:

PR #1121, opened 2026-05-13
fix: reliable auto-start and multi-instance connection support

emiapwil:
## Summary

Fixes two fundamental reliability issues: (1) Unity MCP auto-start never fires on fresh editor launch because EditorPrefs is not initialized during [InitializeOnLoad], and (2) port sharing on Linux due to SO_REUSEADDR being the default, plus multi-instance discovery support.

### HttpAutoStartHandler.cs
- **Root cause**: EditorPrefs returns default values during [InitializeOnLoad]. The static constructor reads AutoStartOnLoad=False and returns before scheduling EditorApplication.delayCall. The delayCall never fires, so the bridge never auto-starts.
- **Fix**: Use EditorApplication.update with retry (up to 120 frames), polling EditorPrefs.GetBool each frame until it returns the correct persisted value.

### StdioBridgeHost.cs
- **Root cause**: ExclusiveAddressUse was gated behind #if UNITY_EDITOR_OSX. On Linux, TcpListener defaults to SO_REUSEADDR, allowing multiple processes to bind the same TCP port. A new Unity instance silently shares port 6400 with an existing one.
- **Fix**: Enable ExclusiveAddressUse = true unconditionally on all platforms.
- **Multi-instance support**: Added pid field to heartbeat. Instance-scoped status file alongside project-scoped file.
- **Stop cleanup**: Deletes both project-scoped and instance-scoped status files.

### PortManager.cs
- **ExclusiveAddressUse**: Unconditional fix in IsPortAvailable().
- **GetPortWithFallback**: Validates stored port is available before returning it.
- **PortConfig**: Added pid field. Instance-scoped port file for multi-instance tracking.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved port selection to avoid reusing ports that are already occupied.
  - Enforced exclusive port binding across platforms to prevent conflicts between running instances.
  - Added process-specific status tracking to improve multi-instance monitoring.
  - Ensured status files are cleaned up when an instance stops.

### DIFF
--- a/.github/workflows/beta-release.yml
+++ b/.github/workflows/beta-release.yml
@@ -13,9 +13,21 @@ on:
       - "MCPForUnity/**"
 
 jobs:
+  unity_tests:
+    name: Unity tests gate
+    if: github.actor != 'github-actions[bot]'
+    uses: ./.github/workflows/unity-tests.yml
+    secrets: inherit
+
+  python_tests:
+    name: Python tests gate
+    if: github.actor != 'github-actions[bot]'
+    uses: ./.github/workflows/python-tests.yml
+
   update_unity_beta_version:
     name: Update Unity package to beta version
     runs-on: ubuntu-latest
+    needs: [unity_tests, python_tests]
     # Avoid running when the workflow's own automation merges the PR
     # created by this workflow (prevents a version-bump loop).
     if: github.actor != 'github-actions[bot]'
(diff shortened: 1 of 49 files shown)


Scriptwonder:
Hi! Sorry for the late response. Can you first retarget the merging to CoplayDev:beta and merge all our recent updates first? I also did some work on Port discovery that might conflict with your change, so it is best to update and run some tests.

jordandunmire97-ai (approved):
The beauty of every incriminate change

Scriptwonder:

jordandunmire97-ai (approved):
thanks

PR #1040, opened 2026-04-06
fix: include UI Toolkit overlays in game_view screenshots with include_image

KennerMiner:
## Summary

Fix the Play Mode `game_view` screenshot path used when `include_image=true` so the returned inline image includes UI Toolkit overlays.

## Problem

When `manage_scene(action="screenshot", capture_source="game_view", include_image=true)` is used without an explicit camera target, the previous implementation captured the image through camera rendering.

That misses UI Toolkit overlays because UI Toolkit is composited after the camera render.

## Change

- Use `ScreenCapture.CaptureScreenshotAsTexture()` for `game_view` screenshots when:
  - `include_image=true`
  - Play Mode is active
  - no explicit camera target is requested
- Keep existing behavior for:
  - explicit camera screenshots
  - non-Play-Mode `include_image` screenshots
  - screenshots where `include_image` is omitted or `false`

## Validation

- Diff is limited to:
  - `MCPForUnity/Editor/Tools/ManageScene.cs`
  - `MCPForUnity/Runtime/Helpers/ScreenshotUtility.cs`
- No package, lockfile, or unrelated tool changes included
- Branch is based on `beta`

## Notes

This PR intentionally keeps the fix narrow to the inline-image screenshot path.

Closes #1039.

## Summary by Sourcery

Update game view screenshot capture to use composited frame grabs for inline images while preserving existing camera-based workflows.

Bug Fixes:
- Ensure Play Mode game_view screenshots with include_image capture UI Toolkit overlays by using composited frame capture instead of camera rendering when no explicit camera is specified.

Enhancements:
- Introduce a composited screenshot helper that leverages ScreenCapture.CaptureScreenshotAsTexture with a camera-based fallback and optional downscaling/base64 encoding for inline images.

## Summary by CodeRabbit

* **New Features**
  * Added a composited screenshot option that captures UI overlays, with optional embedded Base64 image data and automatic downscaling when needed.
  * Screenshot responses now include consistent metadata (path, fullPath, superSize, isAsync, camera/capture source).

* **Bug Fixes**
  * Safer fallback when no explicit camera is set and clearer error messages.
  * Ensures immediate availability of saved screenshots by forcing synchronous asset import.

PR #1132, opened 2026-05-17
fix(screenshot): wait for end-of-frame before composited capture

### LINKED ISSUE #1039: game_view screenshots with include_image miss UI Toolkit overlays
## Summary

`manage_scene(action="screenshot", capture_source="game_view", include_image=true)` can miss UI Toolkit overlays in Play Mode.

## Problem

When `include_image=true` is used without an explicit camera target, the screenshot path captures through camera rendering so it can return image bytes inline in the response.

That misses UI Toolkit content because UI Toolkit is composited after the camera render.

## Expected behavior

When requesting a `game_view` screenshot with `include_image=true` in Play Mode, the returned inline image should include the final composited frame, including UI Toolkit overlays.

## Repro

1. Enter Play Mode in a scene that displays UI Toolkit content in the Game view.
2. Call `manage_scene(action="screenshot", capture_source="game_view", include_image=true)`.
3. Inspect the returned image.
4. Observe that world rendering is present but UI Toolkit overlays are missing.

## Proposed fix

Use composited frame capture for the affected inline-image path, while keeping existing behavior for explicit camera screenshots and other screenshot modes.

## Notes

This appears specific to the `game_view` + `include_image=true` Play Mode path.

KamilDev:
## Description

`manage_camera action=screenshot capture_source=game_view include_image=true` (no explicit camera, play mode) returned fully-transparent PNGs (every pixel RGBA `(0, 0, 0, 0)`). The path was supposed to capture UI Toolkit overlays — #1040 wired it to `ScreenCapture.CaptureScreenshotAsTexture` for that reason — but called the API inline, before the next frame was rendered, so the captured backbuffer was unwritten and the result was always blank.

This change routes the call through a `WaitForEndOfFrame` coroutine and advances the player loop with `EditorApplication.Step()` so the captured texture actually contains the composited frame. Single-call API is preserved.

## Type of Change

## Changes Made

`MCPForUnity/Runtime/Helpers/ScreenshotUtility.cs`:

- Added `CompositedFrameCapturer` `MonoBehaviour` (nested, `#if UNITY_EDITOR`): yields `WaitForEndOfFrame`, then calls `ScreenCapture.CaptureScreenshotAsTexture(SuperSize)` and stores the result in static state.
- Added `CaptureCompositedAfterFrame(int superSize, int timeoutSteps = 5)`: spawns the capturer GameObject, advances the player loop with `EditorApplication.Step()` up to `timeoutSteps` times until the coroutine completes, returns the texture (or `null`, which falls through to existing camera fallback).
- Routed `CaptureComposited`'s editor + play-mode path through `CaptureCompositedAfterFrame`. Edit-mode editor path and non-editor builds retain the previous direct synchronous call.

The public API (`ScreenshotUtility.CaptureComposited` signature, `manage_camera screenshot` tool surface) is unchanged.

Tested locally against `beta` HEAD (`a2a5edf`) with Unity 6000.3.14f1, URP, Windows 11.

**Repro**:

1. Scene with a `UIDocument` using a `PanelSettings` asset.
2. `PanelSettings.clearColor = true`, `colorClearValue = magenta` (unambiguous signal for overlay capture).
3. Enter play mode.
4. `manage_camera action=screenshot capture_source=game_view include_image=true` (no `camera`).

| Path | Before | After |
|---|---|---|
| `manage_camera screenshot` (no camera, `include_image=true`, play mode) | fully-transparent PNG | composited frame including UITK overlay ✅ |
| `manage_camera screenshot camera="Main Camera"` (explicit camera) | camera render only, no overlay (by design) | unchanged ✅ |
| `manage_ui render_ui` (uses its own coroutine path) | works | unchanged ✅ |

No automated test added. The fix's correctness depends on play-mode end-of-frame behavior with a live `UIDocument` — out of scope for the existing edit-mode test fixtures in `TestProjects/UnityMCPTests/Assets/Tests/EditMode/`. Happy to add one if reviewers can point me at the right play-mode test harness pattern in this repo.

## Documentation Updates

This change fixes the behavior of `manage_camera action=screenshot`'s composited path but does not change its parameters, response schema, or add/remove any tool or resource. No documentation updates needed.

## Related Issues

Fixes #1039 (the underlying behavior described in the issue persisted after #1040 — the API choice in #1040 was correct, but the synchronous invocation produced empty captures).

## Additional Notes

- This complements #1040 rather than reverting any of it. #1040's diff is the right API call (`ScreenCapture.CaptureScreenshotAsTexture` instead of camera render), and that line is preserved — just moved inside the coroutine.
- `ManageUI.cs` already contains an equivalent `MCP_ScreenCapturer` MonoBehaviour used by `render_ui`'s two-call pending/ready protocol. A future refactor could share `CompositedFrameCapturer` between both call sites and offer a single-call shape for `render_ui` too. Left out to keep this PR focused.
- `CaptureCompositedAfterFrame` resets stale in-flight state on entry rather than refusing, so a previous failed capture does not poison subsequent attempts.
- `EditorApplication.Step()` is play-mode-only; this is fine because the fixed code path is already gated by `Application.isPlaying`.
- **Edit-mode overlay capture is intentionally not addressed.** `ScreenCapture.CaptureScreenshotAsTexture()` returns `null` when called from a script in edit mode (confirmed empirically), and `EditorApplication.Step()` is a no-op outside play mode — so neither the API nor the spin primitive used by this fix is usable there. Edit-mode `include_image=true` continues to fall through to the existing camera-based path, which excludes UI Toolkit overlays by design. A real edit-mode overlay capture would need a different mechanism (camera RT + `PanelSettings.targetTexture` readback composited manually) — out of scope for this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Editor Play-mode screenshots now reliably include UI overlays and avoid blank captures; previous editor edit-mode and non-editor capture behaviors and camera fallback remain unchanged.

* **Refactor**
  * Play-mode screenshot flow streamlined: editor captures are coordinated to produce composited end-of-frame images more reliably and synchronously.

[Review Change Stack](https://app.coderabbit.ai/change-stack/CoplayDev/unity-mcp/pull/1132?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Scriptwonder:
Hi thanks for the PR! I think its a little over-engineered and this is actually duplication of ManageUI.cs at line 819, so it would be great if you reuse the wheel or call into that wheel. Trim the comments also, keep it under 1-2 lines.

KamilDev:
Done in `a697d09`. Extracted a shared `ScreenshotCapturer` MonoBehaviour to `Runtime/Helpers/ScreenshotUtility.cs` and rewired both call sites (`ManageUI.render_ui`'s two-call protocol and `CaptureComposited`'s sync-spin path). Net -44 lines across the two files. Trimmed comments in the same pass (`0c277af`). Re-tested both paths end-to-end against the magenta UITK overlay repro — no behavior change.

Follow-up nit from CodeRabbit on the refactor (late-callback texture leak after timeout) addressed in `48fbad4`.

Scriptwonder:
Hi, can you keep it updated to the beta so we could run the CI so it would not fail on other Unity versions

KamilDev:
Merged latest `beta` into the branch (`2a7ec426`). Clean merge — none of the 90+ upstream commits touched `ScreenshotUtility.cs` or `ManageUI.cs`, so the diff vs current `beta` is still just the two files (+71 / -37). CI should re-run against the latest matrix.

---

**Update (`6e4302db`):** Spoke too soon. Between that comment and now, beta moved further forward and `4805bc12` landed a compat fix that **did** touch `ScreenshotUtility.cs` — it switched `CaptureComposited` to a reflective `MethodInfo` dispatch so the file compiles when the Screen Capture module is disabled. Re-merged with a real conflict resolution: extracted a single shared `InvokeCaptureScreenshotAsTexture(int)` helper on `ScreenshotUtility` that does the reflective dispatch, and routed both the existing call site and the new `ScreenshotCapturer.Start()` through it. Now consistent with the upstream compat fix. Verified locally — compiles clean, magenta UITK overlay still captured in play mode via `manage_camera screenshot`.

PR #1184, opened 2026-05-31
Serialize expensive editor operations with leases

liuzqk:
## Summary

Add a small cross-process operation lease for expensive editor operations that target the same Unity instance.

- add an atomic file lease shared by MCP server processes for the same Unity instance
- return structured `operation_busy` + `hint=retry` responses when another agent owns the operation window
- guard `refresh_unity` for the full refresh/wait operation
- guard the `run_tests` startup path while preflight and async job dispatch run
- allow reentrant acquisition for the same tool invocation so `run_tests` preflight can still refresh dirty assets without dropping the outer lease
- document the intentionally small scope and non-goals

## Why

This is aimed at the multi-agent workflow where two or more MCP clients point at the same Unity project. In that setup, agents can independently decide to refresh assets, request compilation, or start tests before Unity has updated its own busy state. The result is avoidable contention: duplicate refreshes, overlapping compile waits, or test-start attempts that make the project feel much slower.

The lease gives the server a small coordination layer before those expensive operations reach Unity. It does not serialize normal read-only tools, and it does not try to become a full job queue. Clients still get a retryable response, so they can back off and retry instead of piling more work onto the editor.

## Validation

- `uv run --extra dev python -m pytest tests/integration/test_editor_operation_lease.py tests/integration/test_run_tests_async.py tests/integration/test_refresh_unity_retry_recovery.py -q`
- `git diff --check`

## Release Notes

* **New Features**
  * Multi-agent coordination prevents concurrent operations on the same Editor instance. Multiple agents can now safely interact with the same Unity Editor without operations interfering with each other.
  * Refresh and test operations now return a retry response when already in-progress under another agent.

* **Tests**
  * Added integration tests for operation coordination and busy responses.

* **Documentation**
  * Added operation lease specification document.

### DIFF
--- a/Server/src/services/tools/editor_operation_lease.py
+++ b/Server/src/services/tools/editor_operation_lease.py
@@ -1,0 +1,241 @@
+from __future__ import annotations
+
+import json
+import logging
+import os
+import re
+import time
+import uuid
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+
+from models import MCPResponse
+
+logger = logging.getLogger(__name__)
+
+LEASE_DIR_ENV = "UNITY_MCP_OPERATION_LEASE_DIR"
+LEASE_TTL_ENV = "UNITY_MCP_OPERATION_LEASE_TTL_S"
+DEFAULT_LEASE_TTL_S = 120.0
+_SAFE_KEY_RE = re.compile(r"[^A-Za-z0-9_.@-]+")
+
+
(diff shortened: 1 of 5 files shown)


jordandunmire97-ai (approved):
I hope you see this and do not assume:( i cant do anything if so . It would be okay i have to stay at peace.  Thank you for the evolution of more knowledge that is to prosper.

PR #1175, opened 2026-05-27
fix(stdio): restore 2s socket-release wait to prevent Windows port leak (#1173)

Scriptwonder:
## Summary
- Restore `toWait.Wait(2000)` in `StdioBridgeHost.Stop()` (was reduced 2000 → 500 in PR #787, regressed since 9.7.1)
- Restore matching `stopTask.Wait(2000)` in `StdioBridgeReloadHandler.OnBeforeAssemblyReload`

## Root cause (from #1173)
PR #688 set the 2000ms wait specifically so the `TcpListener` socket could fully release on Windows before the post-reload `Start()` rebinds. PR #787 later reduced this to 500ms without flagging the Windows dependency.

On Windows, 500ms is too short for the OS to free the socket. After the 2nd play-mode entry within a single Editor session (default Enter Play Mode Options, Reload Domain enabled), the post-reload `Start()` hits `AddressAlreadyInUse`, the catch block silently falls back to a new port via `PortManager.DiscoverNewPort()`, and the Python server keeps talking to the orphan on the original port. Scene-graph tools (`find_gameobjects`, `manage_scene get_hierarchy`, `execute_code`) start returning `busy`/timeout indefinitely because OS round-robin sometimes routes calls to the orphaned listener.

Reporter (#1173) bisected to **Reload Domain** as the variable — `Reload Domain OFF` is clean across 4+ play cycles, consistent with the socket-release race.

Verified via `git log -L` that:
- `StdioBridgeHost.cs:400` `Wait(500)` came from commit `79e6c917` (PR #787) which silently reduced PR #688's `Wait(2000)` 
- `StdioBridgeReloadHandler.cs:73` was `500ms` since the file's initial 17cd543f introduction — never had the Windows-safe ceiling

## Scope
Two-line timeout restore; no behavioural change on macOS/Linux (where socket release is fast). The 1.5s extra wait only runs on `Stop()`, which is rare (domain reload / quit), so editor responsiveness is unaffected.

## Test plan
- [ ] Reproduce per #1173: 2× play-mode cycles with Reload Domain ON on Windows; verify port stays on `6400` (no silent shift to `6402`)
- [ ] Verify `find_gameobjects` / `manage_scene get_hierarchy` keep responding after cycle 2
- [ ] Existing `StdioBridgeReconnectTests` continue to pass
- [ ] No regression on macOS (Stop returns within ms on the fast path)

Closes #1173

## Summary by CodeRabbit

* **Bug Fixes**
[...]

Issue #1187, opened 2026-05-31
[Bug]: StdioBridgeHost logs repeated disposed NetworkStream errors on macOS with stdio reconnect/stale clients in 9.7.1


Anywhere-Music-Player:
## What happened?

`MCP-FOR-UNITY` repeatedly logs red Unity Console errors:

```text
MCP-FOR-UNITY: Client handler error: Cannot access a disposed object.
Object name: 'System.Net.Sockets.NetworkStream'.
```

This appears related to #1173, but I am seeing it on macOS / OSXEditor, not Windows. The common part is `MCP for Unity 9.7.1`, `stdio`, and `StdioBridgeHost` handling client/socket lifecycle during reconnects or stale client cleanup.

In my session, Unity is listening on one MCP port, but there are multiple old `mcp-for-unity --transport stdio` Python/uvx processes still running. Their TCP connections to Unity are already `CLOSED`, and Unity then repeatedly logs `ObjectDisposedException` from `HandleClientAsync`.

This does not look like a game/runtime crash. It looks like the stdio bridge closes or loses a client stream, then an async handler continues reading/writing that already disposed `NetworkStream`, and the exception is logged as an error instead of treated as a benign disconnect.

## Environment

- OS: macOS 26.5, build 25F71
- Architecture: arm64
- Unity: 6000.3.16f1
- Unity platform in log: `OSXEditor`
- MCP for Unity package: `9.7.1`
- Python server: `mcpforunityserver==9.7.1`
- Transport: `stdio`
- MCP status file:
  ```json
  {
    "unity_port": 6401,
    "reloading": false,
    "reason": "ready",
    "unity_version": "6000.3.16f1",
    "project_scoped_tools": false
  }
  ```

Package manifest entry:

```json
"com.coplaydev.unity-mcp": "https://github.com/CoplayDev/unity-mcp.git?path=/MCPForUnity#v9.7.1"
```

## Active/stale process evidence

`pgrep -fl "mcp-for-unity --transport stdio|uv tool uvx --from mcpforunityserver|codex.*app-server|mcpbridge"` showed multiple stdio servers/clients alive at the same time:

```text
7380  /opt/homebrew/bin/uv tool uvx --from mcpforunityserver==9.7.1 mcp-for-unity --transport stdio
7449  Python .../mcp-for-unity --transport stdio
69274 /opt/homebrew/bin/uv tool uvx --from mcpforunityserver==9.7.1 mcp-for-unity --transport stdio
[...]

Issue #1164, opened 2026-05-25
HTTP transport: orphaned python server after domain reload (port collision on resume)


malard:
## Summary

When using the HTTP transport, `HttpBridgeReloadHandler.OnBeforeAssemblyReload` calls `transport.ForceStop(TransportMode.Http)` to tear down the python server before a Unity domain reload. In some cases this silently fails to actually kill the python process. After the reload, `ResumeHttpWithRetriesAsync` retries `StartAsync` up to 6 times (0s, 1s, 3s, 5s, 10s, 30s — ~49s total) and every attempt collides with the still-alive python server:

> Cannot start the local HTTP server because port \<N> is already in use by PID(s): \<X>
> MCP For Unity will not terminate unrelated processes. Stop the owning process manually or change the HTTP URL.

Eventually the connection section gives up:

> MCP-FOR-UNITY: Server no longer running; ending orphaned session.

The python process meanwhile is healthy — its log shows the WebSocket close with code 1005 (no status, i.e. the Unity end of the socket disappeared without a proper close handshake) at the moment of domain reload, then it sits idle waiting for a reconnect that never comes because the Editor plugin is trying to spawn a fresh server instead of reattaching.

## Environment

- MCP for Unity 9.7.1
- Unity 6000.3.5f2
- Windows 11
- HTTP transport, configured on a non-default port (reproduced on both 8080 and 8280)

## Repro

1. Configure HTTP transport, start the server, register a plugin session.
2. Trigger a Unity domain reload (any script edit, package import, refresh).
3. After reload, observe: `Cannot start the local HTTP server because port N is already in use by PID(s): X` in the Console, where X is the python PID from before the reload.
4. Confirm the orphan: \`Get-CimInstance Win32_Process -Filter \"ProcessId = X\"\` shows the original \`mcp-for-unity.exe --transport http --http-url http://127.0.0.1:N --pidfile .../Library/MCPForUnity/RunState/mcp_http_N.pid\` invocation, still running.

Python-side log immediately preceding the disconnect:

\`\`\`
2026-05-25 23:44:46,633 - transport.plugin_hub - INFO - Plugin registered: ErtsUnity (...)
2026-05-25 23:44:46,633 - transport.plugin_hub - INFO - Registered 30 tools for session ...
2026-05-25 23:46:36,636 - transport.plugin_hub - INFO - Plugin session ... disconnected (1005)
\`\`\`

## Root cause (as far as I can read it)

[...]

Issue #1130, opened 2026-05-16
StdioBridgeHost: long ExecuteMenuItem runs N times due to broker retry storm under 30s frame timeout


vrxgaminginc:
## Summary

A long, synchronous `ExecuteMenuItem` (longer than `FrameIOTimeoutMs = 30s`) gets executed **N times back-to-back** instead of once, because `StdioBridgeHost.HandleClientAsync` times out the awaiting TCS but does not cancel/remove the queued command, and the broker reconnects and enqueues duplicates that all fire when the original finally completes.

In our repro: a Unity menu item that runs `BuildPipeline.BuildPlayer` for ~6 minutes ran **6 consecutive Android builds** before we manually cancelled. Each was a full ~6-min build.

## Repro

1. Define an `EditorApplication.ExecuteMenuItem`-callable menu that does any synchronous work taking longer than `FrameIOTimeoutMs` (currently 30000 ms). Easy stub: a menu that does `Thread.Sleep(35_000)`. Realistic case: any `BuildPipeline.BuildPlayer` invocation.
2. From an MCP client (e.g. Claude Code), invoke the menu via the `execute_menu_item` tool.
3. Observe in `Editor.log` that the menu item runs N times sequentially, where N equals the number of broker reconnect attempts before it gives up.

The Editor.log shows the same stack repeated for each iteration:

```
<your menu method>
MCPForUnity.Editor.Tools.ExecuteMenuItem:HandleCommand
MCPForUnity.Editor.Tools.CommandRegistry:ExecuteCommand
TransportCommandDispatcher:ProcessCommand → ProcessQueue
StdioBridgeHost:ProcessCommands
UnityEditor.EditorApplication:Internal_CallUpdateFunctions
```

Verified against branch `beta`.

## Root cause

There are two independent timeout paths, both set to the same 30s value (`FrameIOTimeoutMs = 30_000` in `StdioBridgeHost.cs`), and one of them creates the retry-storm.

### Path 1 — outer per-frame TCS timeout in `StdioBridgeHost.HandleClientAsync` (~lines 582-637)

`HandleClientAsync` reads a framed command, enqueues it into `commandQueue` with a `TaskCompletionSource<string>`, then awaits that TCS with a `Task.Delay(FrameIOTimeoutMs)` race.

[...]

PR #1198, opened 2026-06-14
fix(stdio): retry same port on bind race instead of silent fallback (#1173)

### LINKED ISSUE #1173: [Bug]: TcpListener leak after Domain Reload regressed in 9.7.1 on Windows (stdio) — closed by PR #688 / issue #692, but listener-task wait reduced from 2000ms → 500ms and silent port-fallback masks the failure
### What happened?

PR #688 (which closed #692) is structurally present in 9.7.1 - ExclusiveAddressUse=true on Windows, AssemblyReloadEvents.beforeAssemblyReload registered, listener.Server.Dispose() in Stop() - but the listener-task wait was reduced from PR #688's 2000ms back to 500ms, which appears too short on Windows for the socket to fully release. The new bridge then hits AddressAlreadyInUse, silently falls back to a new port via PortManager.DiscoverNewPort(), and the Python server keeps talking to the orphan on the original port. Symptoms match #692 exactly.

After the 2nd play-mode entry within a single Editor session (default Enter Play Mode Options, Reload Domain enabled), scene-graph tools - find_gameobjects, manage_scene get_hierarchy, execute_code — return {"success": false, "error": "busy", "data": {"reason": "compiling", "retry_after_ms": 500}} indefinitely or time out. read_console and manage_editor (play/stop) keep working - consistent with OS round-robin between the orphaned and live listeners.

Verified locally against the installed 9.7.1 package:

MCPForUnity/Editor/Services/Transport/Transports/StdioBridgeHost.cs:343 - ExclusiveAddressUse = true  (matches PR #688)
MCPForUnity/Editor/Services/StdioBridgeReloadHandler.cs:33 - AssemblyReloadEvents.beforeAssemblyReload += OnBeforeAssemblyReload (matches PR #688)
MCPForUnity/Editor/Services/Transport/Transports/StdioBridgeHost.cs:374 - listener?.Server?.Dispose() in Stop()  (matches PR #688)
MCPForUnity/Editor/Services/Transport/Transports/StdioBridgeHost.cs:400 - toWait.Wait(500)  (PR #688 used 2000ms)
MCPForUnity/Editor/Services/StdioBridgeReloadHandler.cs:73 - stopTask.Wait(500) (also benefits from a longer timeout)

Suggested action:

Restore the 2000ms toWait.Wait(...) at StdioBridgeHost.cs:400 from PR #688, and bump stopTask.Wait(500) at StdioBridgeReloadHandler.cs:73 correspondingly. The PR description explicitly called out the longer timeout as part of the Windows fix.

Reconsider the silent port-fallback at StdioBridgeHost.cs:285-305. Failing fast on the configured port would surface the orphan instead of masking it. If the fallback is kept, the Python server needs a discovery mechanism (status file, heartbeat) so it doesn't stay glued to the dead port.

Workaround until fixed: Project Settings > Editor > Enter Play Mode Settings > Reload Domain -> off (no reload -> no socket churn). Confirmed clean across 4+ play cycles in this configuration.

### Reproduction steps

1. Fresh Unity Editor session, scene loaded, default Enter Play Mode Options (Reload Domain enabled).
2. Call manage_editor with action=play from the MCP client -> cycle 1 works, tools respond normally.
3. Call manage_editor with action=stop.
4. Call manage_editor with action=play again -> the bridge log now reports StdioBridgeHost started on port 6402 instead of the configured 6400.
5. Call any scene-graph tool (find_gameobjects, manage_scene get_hierarchy, execute_code) -> returns {"success": false, "error": "busy", "data": {"reason": "compiling", "retry_after_ms": 500}} indefinitely, or times out, until play mode exits.

### Unity version

6000.3.16f1 (Unity 6.3 LTS)

### MCP for Unity package version

9.7.1

### Python server version

9.7.1

### MCP client

Claude Desktop

### Transport

stdio

### OS

Windows

### Relevant logs / console output

```text
Bridge startup after the 2nd play-mode entry (port silently shifted from 6400 -> 6402):

MCP-FOR-UNITY: StdioBridgeHost started on port 6402. (OS=WindowsEditor, server=9.7.1)
Game loaded successfully
Offline earnings: 11330.690890233 bones (14406.472842 seconds away)
Initial state: 138447.458279978 bones, 1.573 BPS, 4 BPT
Game saved successfully
Tool calls from the MCP client (configured port 6400) immediately after the port shift:

find_gameobjects(search_term="EventSystem", search_method="by_name")
  -> MCP error -32001: Request timed out
find_gameobjects(search_term="EventSystem", search_method="by_name")  [retry]
  -> {"success": false, "message": "compiling", "error": "busy",
     "data": {"reason": "compiling", "retry_after_ms": 500}, "hint": "retry"}
manage_scene(action="get_hierarchy", max_depth=2)
  -> {"success": false, "message": "compiling", "error": "busy",
     "data": {"reason": "compiling", "retry_after_ms": 500}}
On the play cycle where the orphan first appeared, the Unity console also logged:

MCP-FOR-UNITY: Client handler error: Cannot access a disposed object.
MCP-FOR-UNITY: Client handler error: Cannot access a disposed object.
MCP-FOR-UNITY: Client handler error: Cannot access a disposed object.
MCP-FOR-UNITY: Client handler error: Cannot access a disposed object.
Bisect — Reload Domain is the variable, Run In Background is not:

Reload Domain	Run In Background	Result
ON		OFF			bug reproduces on cycle 2
ON		ON			bug reproduces on cycle 2
OFF		OFF			4+ play cycles clean, port stays on 6400

manage_editor(action="stop")  -> success
find_gameobjects(search_term="EventSystem")  -> {"instanceIDs": [68324], "totalCount": 1}
```

### Checks

- [x] I searched existing issues and did not find a duplicate
- [x] I included logs / steps to reproduce

Scriptwonder:
## Summary
After a domain reload, a transient `AddressAlreadyInUse` on the configured stdio port no longer triggers a silent permanent port switch that strands the Python client on the orphan listener (the #1173 failure). The bridge now retries the **same** port and only falls back to a new port — loudly — when the port stays busy past a short window.

## Root cause (#1173)
`Stop()`/`Dispose()` run on the main thread during `beforeAssemblyReload`, but the OS frees the bound port a few hundred ms later (slower on Windows/macOS). When the post-reload `Start()` rebinds too soon, `listener.Start()` throws `AddressAlreadyInUse`. The previous `catch` silently switched to a new port via `DiscoverNewPort()`, but the Python client stays pinned to the configured port → it talks to the orphan and returns `busy`/timeout indefinitely.

## What changed
- **`StdioBridgeHost.Start()`** — on `AddressAlreadyInUse`, keep the configured port and fail the attempt **without blocking**. The reload handler's async resume schedule (0/1/3/5/10/30s) and the editor-idle retry re-invoke `Start()` on the same port within ~1s, by which point the OS has released it. Only after the port stays busy past `PortManager.BusyPortFallbackWindowSeconds` (3s) does it fall back to a new port — via `McpLog.Warn` instead of the previous debug-only switch. The retry path also re-arms the editor-idle loop so a direct `Start()` caller (e.g. `StartAutoConnect`) can't be left permanently stopped.
- **`PortManager`** — `ShouldAbandonBusyPort(busyForSeconds)` policy + `BusyPortFallbackWindowSeconds` constant.
- **`HandleClientAsync`** — classify `ObjectDisposedException` from a torn-down `NetworkStream` as a benign disconnect (fixes the red Console spam in #1187).
- **`PortManagerTests`** — cover the retry-vs-fallback policy threshold.

## Why not just restore the 2000ms wait (the closed #1175 approach)
Bumping `Stop()`'s wait back to 2000ms re-introduces the backgrounded-Unity stall that #787 deliberately fixed by reducing it to 500ms, and it doesn't address the silent port fallback that is the actual masking layer. This PR keeps the 500ms wait untouched, adds **zero** new main-thread blocking, and is platform-agnostic (the race also reproduces on macOS, per the #1173 thread).

## Test results

**CI — full matrix green** (`safe-to-test` + `full-matrix`):

| Unity | EditMode |
|---|---|
| 2021.3.45f2 | :white_check_mark: pass |
| 2022.3.62f1 | :white_check_mark: pass |
| 6000.0.75f1 | :white_check_mark: pass |
| 6000.4.8f1 | :white_check_mark: pass |

UnityMCPTests EditMode on 2021.3.45f2: **886 passed, 0 failed, 46 skipped** (all `Explicit` side-effect / URP-gated `Inconclusive`).

**Live validation** — the fix loaded into a running Unity **2021.3.45f2** editor:
- Fix code loads + policy correct in the real runtime: `BusyPortFallbackWindowSeconds = 3s`; `ShouldAbandonBusyPort(1.0s) = false` (keep port), `ShouldAbandonBusyPort(3.5s) = true` (fall back).
- End-to-end retry (real `TcpListener` + real `PortManager`): an orphan holding a port triggers `AddressAlreadyInUse`; within the 3s window the port is kept; after the orphan releases, the **same** port rebinds — no silent switch.
- A real domain reload tore down and rebound the bridge — it **recovered** and stayed on the configured port `6400` (no silent shift).
- **#1187:** after reload churn, **0** red `Cannot access a disposed object` errors.

**A/B — same runtime, same orphan-on-`6543` scenario, before vs after the change:**

| port 6543 held → `AddressAlreadyInUse` | Before (no fix) | After (fix) |
|---|---|---|
| same-port retry policy | absent | present (3s window) |
| on conflict | **switches away to 6544** (silent — the #1173 bug) | **keeps 6543**, retries; rebinds same port on release |

## Related issues & regression sweep
- **Closes #1173** — validated above (Before switches the port away; After keeps it).
- **#1187** (disposed `NetworkStream` red errors) — primary symptom fixed by the benign `ObjectDisposedException` reclassification (0 red errors observed under reload churn). The secondary "clean up stale stdio processes" ask is largely client-side and out of scope here.
- **#1164** (HTTP transport: orphaned python server / port collision on resume) — *similar surface, different root cause*: the HTTP path has a Python-server respawn-vs-reattach problem in `HttpBridgeReloadHandler` (`ForceStop` may not kill the process, then `ResumeHttpWithRetriesAsync` spawns a fresh one that collides), not stdio's socket-release timing. **This PR does not address it** (stdio-only) and the same-port-retry approach wouldn't fix it — tracked separately.
- Diff is **stdio-only** (3 files). It does not touch the HTTP transport (#1164), the command broker / `ExecuteMenuItem` retry path (#1130), stale-worker handling (#1090), or multi-instance routing (#1023) — **no regressions introduced** in those areas.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TCP port fallback handling by retrying the configured port for up to **3 seconds** when it’s temporarily busy, only switching to a newly discovered port after the window elapses.
  * Added clearer “port busy” signaling during reload-related race conditions and reduced log noise by treating expected disposal events as benign.
* **Tests**
  * Added unit tests covering the abandon-vs-keep retry behavior within and beyond the 3-second window.

### DIFF
--- a/MCPForUnity/Editor/Helpers/PortManager.cs
+++ b/MCPForUnity/Editor/Helpers/PortManager.cs
@@ -65,6 +65,25 @@ namespace MCPForUnity.Editor.Helpers
         }
 
         /// <summary>
+        /// How long the configured port may keep returning AddressAlreadyInUse before the
+        /// bridge gives up on it and discovers a new one. The common cause of a transient
+        /// conflict is our own previous listener whose OS socket has not been released yet
+        /// after a domain reload (slower on Windows/macOS); this window covers that race so
+        /// the bridge keeps the configured port instead of stranding the client on the
+        /// orphan (#1173).
+        /// </summary>
+        public const double BusyPortFallbackWindowSeconds = 3.0;
+
+        /// <summary>
+        /// Decide whether a configured port that keeps reporting AddressAlreadyInUse should be
+        /// abandoned for a freshly discovered port. Returns false (keep retrying the same port)
+        /// until it has been continuously busy for <see cref="BusyPortFallbackWindowSeconds"/>,
+        /// which distinguishes a domain-reload socket-release race from a foreign occupant (#1173).
+        /// </summary>
+        public static bool ShouldAbandonBusyPort(double busyForSeconds)
+            => busyForSeconds >= BusyPortFallbackWindowSeconds;
+
+        /// <summary>
         /// Persist a user-selected port and return the value actually stored.
         /// If <paramref name="port"/> is unavailable, the next available port is chosen instead.
         /// </summary>
--- a/MCPForUnity/Editor/Services/Transport/Transports/StdioBridgeHost.cs
+++ b/MCPForUnity/Editor/Services/Transport/Transports/StdioBridgeHost.cs
@@ -46,6 +46,9 @@ namespace MCPForUnity.Editor.Services.Transport.Transports
         private static bool isStarting = false;
         private static double nextStartAt = 0.0f;
         private static double nextHeartbeatAt = 0.0f;
+        // EditorApplication.timeSinceStartup of the first AddressAlreadyInUse on the configured
+        // port; 0 when the port binds cleanly. Drives the same-port retry window (#1173).
+        private static double _portBusySince = 0.0;
         private static int heartbeatSeq = 0;
         private static Dictionary<string, QueuedCommand> commandQueue = new();
         private static int mainThreadId;
@@ -284,10 +287,36 @@ namespace MCPForUnity.Editor.Services.Transport.Transports
                     }
                     catch (SocketException se) when (se.SocketErrorCode == SocketError.AddressAlreadyInUse)
                     {
-                        // Port is busy. Try switching to a new port once; if that also fails,
-                        // let the reload handler retry with async backoff instead of blocking here.
+                        // The configured port is busy. The usual cause is our own previous listener
+                        // whose OS socket has not been released yet after a domain reload: Stop()/
+                        // Dispose runs on the main thread, but the kernel frees the bound port a few
+                        // hundred ms later (longer on Windows/macOS).
+                        //
+                        // Do NOT silently switch to a new port on the first conflict — the Python
+                        // client stays pinned to the configured port and ends up talking to the
+                        // orphan, returning busy/timeout forever (#1173). Keep the configured port
+                        // and fail this attempt WITHOUT blocking: the reload handler's async resume
+                        // schedule and the editor-idle retry re-invoke Start() on the same port within
+                        // ~1s, by which point the OS has released it. Only after the port stays busy
+                        // past the fallback window do we treat it as a foreign occupant and switch.
+                        double now = EditorApplication.timeSinceStartup;
+                        // Start a fresh window on the first conflict, or if a stale timestamp
+                        // survived a long idle gap (a real reload + retry resolves in seconds).
+                        if (_portBusySince <= 0.0 || (now - _portBusySince) > 60.0) _portBusySince = now;
+
+                        if (!PortManager.ShouldAbandonBusyPort(now - _portBusySince))
+                        {
+                            try { listener?.Server?.Dispose(); } catch { }
+                            listener = null;
+                            McpLog.Warn($"Port {currentUnityPort} not released yet after reload; retrying same port.");
+                            WriteHeartbeat(true, "port_busy");
+                            nextStartAt = now + 0.3; // throttle the editor-idle retry loop
+                            return;
+                        }
+
                         int oldPort = currentUnityPort;
                         currentUnityPort = PortManager.DiscoverNewPort();
+                        _portBusySince = 0.0;
 
                         try
                         {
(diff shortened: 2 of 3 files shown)


Scriptwonder:
Thanks for the review — addressed in b0fefe4b:

- **CodeRabbit (early-return retry arming):** valid — the previous early-return assumed `EnsureStartedOnEditorIdle` was already hooked, which isn't guaranteed for direct `Start()` callers like `StartAutoConnect`. The retry path now arms the editor-idle loop itself (`if (!ensureUpdateHooked) { … EditorApplication.update += EnsureStartedOnEditorIdle; }`), so a transient `AddressAlreadyInUse` can never leave the bridge permanently stopped.
- **Copilot (listener cleanup):** now calls `listener.Stop()` before `Server.Dispose()`, matching `Stop()`.
- **Copilot (magic number):** `60.0` is now `PortBusyStaleResetSeconds`.

Issue #1207, opened 2026-06-15
macOS / HTTP transport: bridge WebSocket repeatedly drops (close 1005) on reload/test boundaries → no_unity_session


Syntaf:
Hey all, i've been working with the unity MCP for a few months now (love it) but have recently started to run into frequent session disconnects on a new project of mine. Creating an issue here as I continue to look into things, but I may also open a PR that proposes extending the wait time before determining that no unity plugin exists.

**Note:** The investigation below was done by my claude code, and verified by me before posting. Everything above this note is strictly my own voice, everything below this note is Claude's investigation & analysis.

> Likely the macOS counterpart of #691 (Windows-only, `WinError 64`). #691 is closed, but the
> same *class* of failure — the MCP server losing its link to the Unity bridge — reproduces
> consistently on macOS with HTTP transport. The macOS signature differs (close code **1005**, no
> `WinError`/`Accept failed`/`Task exception`), so this may or may not be the same.

## Environment
- **OS:** macOS 14.5 (Darwin 23.5.0), Apple Silicon
- **Unity:** 6000.0.44f1
- **MCP for Unity bridge:** `com.coplaydev.unity-mcp` **v9.7.1** (UPM git `#main`, commit `efaf786`)
- **MCP server:** `mcpforunityserver==9.7.1` via `uvx`, **HTTP transport** on `127.0.0.1:8080`, `--project-scoped-tools`
- **Client:** Claude Code (HTTP MCP, `{"type":"http","url":"http://127.0.0.1:8080/mcp"}`)
- **Python:** 3.14.0 · **uv:** 0.6.6

## Symptom
The Unity plugin's WebSocket to the server's PluginHub disconnects **every few minutes** under active workload, **always with close code 1005** (abrupt / no close frame), **clustered on domain-reload and PlayMode `run_tests` boundaries**. Two outcomes, same 1005 cause:

- **CHURN (usually harmless):** the bridge tears down and resumes across the reload — `[HTTP Reload] Resume attempt 1/6` → `Resume succeeded on attempt 1`. Heals in ~1 attempt.
- **WEDGE (the painful bug):** the reconnect does **not** complete within the server's 20 s window → `No Unity plugin reconnected within 20.00s` → every tool call returns `no_unity_session` and `refresh_unity` times out after 60 s, requiring manual Unity refocus / MCP restart.

[...]

Issue #891, opened 2026-03-08
Mcp would stuck if Unity pops a reload window. I have to click 'reload' mannually to make the mcp work again.


hatrd:
I'm not watching it all the time, so the stuck is really unacceptable.

<img width="391" height="159" alt="Image" src="https://github.com/user-attachments/assets/17bfb9e3-f803-45dc-b320-2b8bb90c5445" />

<img width="729" height="660" alt="Image" src="https://github.com/user-attachments/assets/06c8ade4-0fea-4fde-b92c-e72f27ed8e9d" />

I have to click reload mannually and the mcp would work again.

unity mcp version: v9.4.7

abcdhdev:
Yes I get the same problem with Codex.

Scriptwonder:
@karaok1 @hatrd Have you turned on (Preferences > General > Auto Refresh) option? Unity-MCP needs an extensive domain reload but we don't want to force that.

Leuconoe:
<img width="296" height="89" alt="Image" src="https://github.com/user-attachments/assets/30e2bd64-61ae-40ff-bc29-d43639fa593b" />

Issue #657, opened 2026-01-31
Make domain reload detection more deterministic (reduce 30s timeout)


dsarno:
## Summary

Currently, when Unity performs a domain reload (e.g., after running tests or compiling scripts), the MCP connection drops. The server waits up to **30 seconds** for Unity to reconnect before failing commands.

This 30-second timeout is conservative to handle slow domain reloads, but it can cause unnecessary delays in agentic workflows where domain reloads happen frequently.

## Current Behavior

- `UNITY_MCP_SESSION_RESOLVE_MAX_WAIT_S` defaults to 30 seconds
- `UNITY_MCP_RELOAD_MAX_WAIT_S` defaults to 30 seconds  
- Server blindly waits up to 30s when no Unity session is available

## Proposed Improvement

Make detection more deterministic by:

1. **Detecting Unity's actual reload state** rather than blindly waiting
   - Check the status file for "reloading" state
   - Monitor for connection attempts during the wait period
   - Use exponential backoff with early exit when connection is established

2. **Shorter default with smart detection**
   - Default to ~5-10 seconds for normal operations
   - Extend only when domain reload is detected (via status file or heuristics)

3. **Different timeouts for different contexts**
   - Post-test-suite: longer timeout expected
   - Normal commands: shorter timeout preferred

## Related

- Fixed in PR #656 (increased from 2s to 30s to handle domain reloads)
- Issue #654 (dispatcher becomes unreachable after test suite)

## Files Involved

- `Server/src/transport/plugin_hub.py` - `UNITY_MCP_SESSION_RESOLVE_MAX_WAIT_S`
- `Server/src/transport/legacy/unity_connection.py` - `UNITY_MCP_RELOAD_MAX_WAIT_S`

## Notes

See TODO comment in `unity_connection.py` around line 769 for context.

dsarno:
Think maybe don't need this after all.

PR #1217, opened 2026-06-27
fix(stdio): recover from a reload-orphaned socket instead of hanging

mertakdut:
## Description
Every script recompile makes Unity do a domain reload, which tears down and rebuilds the bridge. The Unity process itself keeps running, so the OS never resets the old loopback connection, and the Python client is left holding a socket that still looks connected but is actually dead. The old code kept reusing that socket: each recv waited the full connection_timeout, and those waits piled up across retries with no overall limit, so a tool call could hang for several minutes. I ran into this with refresh_unity, run_tests, and get_test_job stuck for 10+ minutes at a time.

The fix does two things. When the client sees that Unity is reloading, it drops the stale socket so the next command reconnects to the fresh bridge instead of writing into the dead one. And it puts a single time limit on the whole command, so retries can't stack up forever.

## Type of Change

- [X] Bug fix (non-breaking change that fixes an issue)
- [ ] Documentation update
- [ ] Test update

## Changes Made
1. Server/src/transport/legacy/unity_connection.py
  - Reconnect-on-reload: when the preflight status reports reloading, drop the socket (under _io_lock, so it is never closed under a concurrent send/recv) so the next command reconnects to the rebuilt bridge instead of reusing a half-open one.
  - Per-command total deadline: new optional deadline arg on send_command, derived from command_total_timeout and enforced at the top of the attempt loop; threaded through send_command_with_retry (one shared deadline across the reload-wait loop) so the whole call has a single ceiling rather than one per inner send.
2. Server/src/core/config.py: new command_total_timeout (default 90s).
3. Server/tests/integration/test_connection_deadline.py: deterministic regression tests using a fake bridge that completes the handshake then goes silent.

- Unity version(s) tested: 6000.5.0f1 (Unity 6.5)
- Package source used (`#beta`, `#main`, tag, branch, or `file:`): Python server run from local Server/ source (uvx --from <path>); the Unity C# package was unchanged (beta).
- Resolved commit hash from `Packages/packages-lock.json` (if using a Git package URL): N/A -- this PR changes only the Python server; no Unity C# package change.

- [X] Python tests (`cd Server && uv run pytest tests/ -v`)
- [ ] Unity EditMode tests
- [ ] Unity PlayMode tests
- [ ] Package import/compile check

## Documentation Updates

## Related Issues
Relates to #891 (MCP gets stuck after a Unity reload until manually nudged) and #657 (more deterministic / bounded reload-wait in unity_connection.py). The same reload/test-boundary connection-drop pattern also affects the HTTP transport (e.g. #1207, #1164) via a different code path; this PR addresses the stdio client only.

## Additional Notes
1. This change is Python-only (no C# / Unity API change), so the Unity EditMode/PlayMode and package-import/compile checks are not applicable.
3. The bug is timing/state-dependent and does not reproduce on demand, so the new test reproduces it deterministically: a fake bridge that handshakes then goes silent drives the real connection code, failing (unbounded) before the fix and bounded after. It was also verified live on Unity 6.5 against the exact workflow that originally hung (repeated edit -> force refresh -> run_tests / get_test_job): previously wedged for minutes; now every call returns within ~10s and reconnects cleanly across each reload.
4. Related but intentionally out of scope: after heavy reload churn the Unity-side bridge sometimes fails to resume (its status file is deleted on stop and not recreated, leaving the bridge down until the editor restarts). That is a separate, bridge-side issue.

## Summary by CodeRabbit

* **New Features**
  * Added a `command_total_timeout` configuration to cap how long commands can run across retries.

* **Bug Fixes**
  * Improved Unity command sending so timeouts are enforced consistently, even after a reload leaves the socket half-open.
  * Strengthened reload handling to prevent concurrent send/receive races and to ensure timed-out sockets are reset.

* **Tests**
  * Added integration coverage for silent/half-open TCP behavior and reload-related retry timing and error reasons.

### DIFF
--- a/Server/src/core/config.py
+++ b/Server/src/core/config.py
@@ -32,6 +32,8 @@ class ServerConfig:
 
     # Connection settings
     connection_timeout: float = 30.0
+    # Hard ceiling on a command's total time across all retries (wedged-socket guard).
+    command_total_timeout: float = 90.0
     buffer_size: int = 16 * 1024 * 1024  # 16MB buffer
 
     # STDIO framing behaviour
--- a/Server/src/transport/legacy/unity_connection.py
+++ b/Server/src/transport/legacy/unity_connection.py
@@ -50,7 +50,7 @@ class UnityConnection:
         except OSError as exc:
             logger.debug(f"Unable to set TCP_NODELAY: {exc}")
 
-    def connect(self) -> bool:
+    def connect(self, connect_timeout: float | None = None) -> bool:
         """Establish a connection to the Unity Editor."""
         if self.sock:
             return True
@@ -59,8 +59,9 @@ class UnityConnection:
                 return True
             try:
                 # Bounded connect to avoid indefinite blocking
-                connect_timeout = float(
-                    getattr(config, "connection_timeout", 1.0))
+                if connect_timeout is None:
+                    connect_timeout = float(
+                        getattr(config, "connection_timeout", 1.0))
                 # We trust config.unity_host (default 127.0.0.1) but future improvements
                 # could dynamically prefer 'localhost' depending on OS resolver behavior.
                 self.sock = socket.create_connection(
@@ -265,7 +266,13 @@ class UnityConnection:
             logger.error(f"Error during receive: {str(e)}")
             raise
 
-    def send_command(self, command_type: str, params: dict[str, Any] = None, max_attempts: int | None = None, deadline: float | None = None) -> dict[str, Any]:
+    def _cap_to_deadline(self, timeout: float, deadline: float | None, floor: float = 0.05) -> float:
+        """Shrink a blocking timeout to whatever budget remains before the deadline."""
+        if deadline is None:
+            return timeout
+        return max(floor, min(timeout, deadline - time.monotonic()))
+
+    def send_command(self, command_type: str, params: dict[str, Any] | None = None, max_attempts: int | None = None, deadline: float | None = None) -> dict[str, Any]:
         """Send a command with retry/backoff and port rediscovery. Pings only when requested.
 
         Args:
@@ -320,8 +327,6 @@ class UnityConnection:
                 logger.debug(f"Preflight status check failed: {exc}")
                 return None
 
-        last_short_timeout = None
-
         # Extract hash suffix from instance id (e.g., Project@hash)
         target_hash: str | None = None
         if self.instance_id and '@' in self.instance_id:
@@ -356,7 +361,7 @@ class UnityConnection:
                 self._ensure_live_connection()
                 # Ensure connected (handshake occurs within connect())
                 t_conn_start = time.time()
-                if not self.sock and not self.connect():
+                if not self.sock and not self.connect(self._cap_to_deadline(config.connection_timeout, deadline)):
                     raise ConnectionError("Could not connect to Unity")
                 logger.info("[TIMING-STDIO] connect took %.3fs command=%s", time.time() - t_conn_start, command_type)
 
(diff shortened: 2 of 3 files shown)


mertakdut:
> 🧹 Nitpick comments (1)
> 
> > Server/src/transport/legacy/unity_connection.py (1)> `392-402`: _🩺 Stability & Availability_ | _🔵 Trivial_ | _🏗️ Heavy lift_
> > > **Thread the shared deadline through `_read_exact`**
> > > `receive_full_response()` can make many blocking `recv()` calls under one socket timeout. A slow peer that keeps dribbling bytes can keep the command alive past `command_total_timeout`, so the “hard ceiling” only applies to the first read. Pass the deadline into `_read_exact` and cap/check before each `recv()`.
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against current code. Fix only still-valid issues, skip the
> > > rest with a brief reason, keep changes minimal, and validate.
> > > 
> > > In `@Server/src/transport/legacy/unity_connection.py` around lines 392 - 402, The
> > > shared command deadline is only being applied before the first socket read in
> > > receive_full_response() / _read_exact, so a slow peer can still extend past the
> > > total timeout. Thread the deadline into _read_exact and, using the existing
> > > helpers in UnityConnection, cap the recv timeout and re-check the deadline
> > > before each recv() loop iteration so every blocking read respects the same hard
> > > ceiling.
> > > ```
> 
> 🤖 Prompt for all review comments with AI agents
> 
> ℹ️ Review info

@coderabbitai true in principle, but it can't really happen here. The peer is the Unity bridge on 127.0.0.1 and it sends each response in a single sendall, so over loopback _read_exact reads it in a couple of passes and never gets close to the timeout. The case that actually matters (socket wedged, no bytes at all) is already capped on the first recv, and a slow-but-alive Unity is covered by the heartbeat window. I'll leave _read_exact as-is for now, happy to revisit if this ever runs over a real network.

fs-paulrahme:
> > 🧹 Nitpick comments (1)
> > > Server/src/transport/legacy/unity_connection.py (1)> `392-402`: _🩺 Stability & Availability_ | _🔵 Trivial_ | _🏗️ Heavy lift_
> > > > **Thread the shared deadline through `_read_exact`**
> > > > `receive_full_response()` can make many blocking `recv()` calls under one socket timeout. A slow peer that keeps dribbling bytes can keep the command alive past `command_total_timeout`, so the “hard ceiling” only applies to the first read. Pass the deadline into `_read_exact` and cap/check before each `recv()`.
> > > > 🤖 Prompt for AI Agents
> > > > ```
> > > > Verify each finding against current code. Fix only still-valid issues, skip the
> > > > rest with a brief reason, keep changes minimal, and validate.
> > > > 
> > > > In `@Server/src/transport/legacy/unity_connection.py` around lines 392 - 402, The
> > > > shared command deadline is only being applied before the first socket read in
> > > > receive_full_response() / _read_exact, so a slow peer can still extend past the
> > > > total timeout. Thread the deadline into _read_exact and, using the existing
> > > > helpers in UnityConnection, cap the recv timeout and re-check the deadline
> > > > before each recv() loop iteration so every blocking read respects the same hard
> > > > ceiling.
> > > > ```
> > 
> > 
> > 🤖 Prompt for all review comments with AI agents
> > ℹ️ Review info
> 
> @coderabbitai true in principle, but it can't really happen here. The peer is the Unity bridge on 127.0.0.1 and it sends each response in a single sendall, so over loopback _read_exact reads it in a couple of passes and never gets close to the timeout. The case that actually matters (socket wedged, no bytes at all) is already capped on the first recv, and a slow-but-alive Unity is covered by the heartbeat window. I'll leave _read_exact as-is for now, happy to revisit if this ever runs over a real network.

We've been chasing intermittent infinite "Reloading Domain" hangs (sometimes script recompile, sometimes and play-mode entry). Dumps of the not-responding `Unity.exe`, with an ongoing Claude session from the last 2-3 weeks to help with pattern matching and narrowing it down.  I've re-edited most of this response myself, so hope it doesn't still read too AI generated, or apologies if you're already further than us 😄 Just sharing in case it surfaces any further insight...

Our `.exe` dumps consistently show the main thread stuck in `mono_domain_try_unload`, waiting on a thread that never exits — and the "Thread Pool I/O Selector" thread blocked in `ws2_32!select` (via `mono_poll`) on a `127.0.0.1` bridge socket. That's Unity's documented domain-reload-hang cause: a thread in a blocking socket call can't be terminated during reload, so `mono_domain_try_unload` never completes.

Thus the editor-side bridge's pending socket receive isn't cancelled/closed before the domain unloads.

The Python-side recovery can help recover from / work around this in the meantime, but it looks like the cause goes deeper into editor code: on `AssemblyReloadEvents.beforeAssemblyReload` (and `EditorApplication.quitting`), close the bridge socket and stop/join its receive thread so the reload completes cleanly. This PR's client-side recovery handles the client holding a dead socket; the editor still needs to tear down its end on reload to avoid the hard hang.

jordandunmire97-ai (approved):
Please understand I am just confirming these myself.  It's hard to move my hands . I see they do need to revised before merge this is just easier for me in this context thank you for everything

Scriptwonder:
Thanks for the PR! Take some time for me to test this, I agree with @fs-paulrahme that this might only solve half of the problem, but at least one step forward.

PR #1073, opened 2026-04-24
Add LAN HTTP transport mode

ff6330858:
## Type of Change

Save your change type
- Bug fix (non-breaking change that fixes an issue)
- New feature (non-breaking change that adds functionality)
- Breaking change (fix or feature that would cause existing functionality to change)
- Documentation update
- Refactoring (no functional changes)
- Test update

## Documentation Updates

- [ ] I have added/removed/modified tools or resources
- [ ] If yes, I have updated all documentation files using:
  - [ ] The LLM prompt at `tools/UPDATE_DOCS_PROMPT.md` (recommended)
  - [ ] Manual updates following the guide at `tools/UPDATE_DOCS.md`

## Summary by CodeRabbit

* **New Features**
  * Added Local Area Network (LAN) HTTP transport option for server connections.
  * Separate LAN public and bind URL settings with distinct persisted values.
  * UI now shows a prominent warning when LAN HTTP is enabled.

* **Bug Fixes / Improvements**
  * LAN-aware validation and launch checks to enforce secure URL requirements.
  * Transport selection, status reporting, and server start/stop behavior now respect the LAN scope.

Issue #549, opened 2026-01-12
`EditorApplication.isCompiling` False Positive in Play Mode


hui1993:
# Bug Report: `EditorApplication.isCompiling` False Positive in Play Mode

## Summary
`EditorApplication.isCompiling` returns `true` even when Unity is not actually compiling, causing MCP tools to incorrectly return `"error": "busy", "reason": "compiling"` in Play mode.

## Environment
- Unity Version: 2022.3.x
- MCP for Unity Version: Latest (git)
- OS: macOS

## Steps to Reproduce
1. Enter Play mode in Unity Editor
2. Use MCP client (Cursor) to call tools like `find_gameobjects` or `manage_scene`
3. Observe intermittent `"error": "busy", "reason": "compiling"` errors even though Unity is not compiling

## Root Cause Analysis
Through debugging, we found that `EditorApplication.isCompiling` can return `true` in Play mode even when `CompilationPipeline.isCompiling` (via reflection) returns `false`.

### Debug Evidence
```json
// Unity reports compiling but actually not compiling:
{
  "editorAppIsCompiling": true,      // ← False positive!
  "pipelineIsCompiling": false,      // ← Actually not compiling
  "possibleFalsePositive": true
}
```

## Affected Code
`Editor/Services/EditorStateCache.cs` uses `EditorApplication.isCompiling` directly:

```csharp
// Line 292
bool isCompiling = EditorApplication.isCompiling;
```

## Proposed Fix
Add a helper method that double-checks with `CompilationPipeline.isCompiling` in Play mode:

```csharp
private static bool GetActualIsCompiling()
{
    // First check EditorApplication.isCompiling
    if (!EditorApplication.isCompiling)
    {
        return false;
    }
    
    // In Play mode, EditorApplication.isCompiling can have false positives
    // Double-check with CompilationPipeline.isCompiling via reflection
    if (EditorApplication.isPlaying)
    {
        try
        {
            Type pipeline = Type.GetType("UnityEditor.Compilation.CompilationPipeline, UnityEditor");
            var prop = pipeline?.GetProperty("isCompiling", BindingFlags.Public | BindingFlags.Static);
            if (prop != null)
            {
                return (bool)prop.GetValue(null);
            }
        }
        catch { }
    }
    
    return EditorApplication.isCompiling;
}
```

Then replace:
```csharp
bool isCompiling = EditorApplication.isCompiling;
```
with:
```csharp
bool isCompiling = GetActualIsCompiling();
```

[...]

PR #1234, opened 2026-07-04
fix(editor): survive domain reloads in HTTP auto-start and reload-resume (#1229)

### LINKED ISSUE #1229: [Bug] v10.0.0: HTTP auto-start and reload-resume both lost to delayCall/domain-reload race (Windows)
### Environment
- MCP for Unity **v10.0.0** (git `#v10.0.0`), HTTP Local transport, "Auto-Start on Editor Load" enabled
- Unity **6000.3.9f1**, Windows 11, DX12
- Client: Claude Code via MCP

### Symptoms
1. Opening the project with any pending script compilation: the bridge **never starts** for the entire editor session. MCP clients report `No Unity Editor instances found` / `no_unity_session`.
2. With the bridge running, importing an asset package that contains scripts (multi-pass compile, e.g. Easy Save): the bridge dies at the reload and **never resumes**.

Both reproduce ~100% under those conditions. Possibly related in symptom (but a distinct mechanism) to #1207.

### Root cause 1 - HttpAutoStartHandler.cs (cold start)
The static ctor gates on a SessionState latch that persists across domain reloads, sets the latch **before** the work happens, then schedules the actual start via `EditorApplication.delayCall`:

- `if (SessionState.GetBool(SessionInitKey, false)) return;`  (gate)
- `SessionState.SetBool(SessionInitKey, true);`  (latch set eagerly)
- `EditorApplication.delayCall += OnEditorReady;`  (deferred work)

If a domain reload lands before the delayCall fires - which always happens when editor startup includes a recompile - the pending delayCall is wiped, and on the next domain load the latch short-circuits the constructor. Auto-start is silently dead for the whole session.

### Root cause 2 - HttpBridgeReloadHandler.cs (reload resume)
`OnAfterAssemblyReload` **consumes** the one-shot resume flag (`EditorPrefs.DeleteKey(ResumeHttpAfterReload)`) and then, when the editor is still compiling, defers the resume via `EditorApplication.delayCall`. Multi-pass compiles trigger a second reload that wipes the delayCall - with the flag already deleted, the resume is lost permanently.

### Suggested fixes
- Set the session latch (and delete the resume flag) only **after** a successful start, not before the deferred work runs.
- Avoid `EditorApplication.delayCall` for anything that must survive a domain reload; an `[InitializeOnLoad]` + `EditorApplication.update` tick re-registers every reload and cannot be wiped, or re-schedule from `afterAssemblyReload` each time.
- Alternatively: on every domain load, if prefs say auto-start and the bridge is not running, just start it - idempotent and race-free.

### Workaround we are shipping
An editor-side watchdog that ticks on `EditorApplication.update`, waits 10s of editor idle (so the package's own retries win when they work), and re-invokes `HttpAutoStartHandler.OnEditorReady` via reflection when the bridge is down; it disarms via a plain static (re-arms every reload). Happy to PR a proper fix if useful.

Found and diagnosed with Claude Code against the v10.0.0 package source; report drafted with AI assistance and human-verified.

Scriptwonder:
Fixes #1229. Full credit to @jtpistella for the precise two-part root-cause diagnosis in the issue — both mechanisms verified line-for-line before writing this.

## Root causes (verified at beta HEAD)

1. **Auto-start dead for the whole session** when startup includes a compile: `HttpAutoStartHandler`'s cctor set the reload-surviving `SessionState` latch *before* registering the `EditorApplication.delayCall` — a domain reload wipes the pending delegate, and the latch then short-circuits every later domain load.
2. **Reload-resume permanently lost at multi-pass compiles** (e.g. importing a script-bearing asset package): `HttpBridgeReloadHandler` consumed the one-shot resume flag before the deferred (`delayCall`) resume ran, and the next `beforeAssemblyReload` deleted it again because the bridge wasn't running.

## Fix

- **Update ticks instead of `delayCall`**: the `[InitializeOnLoad]` cctor re-arms an `EditorApplication.update` tick on every domain load, and the latch is written only when the deferred work actually dispatches. Services-not-ready frames retry (bounded per domain); the common case (auto-start off, nothing pending) skips the subscription entirely.
- **Resume flag moved EditorPrefs → SessionState** and it now persists until the resume *succeeds, is cancelled, or exhausts its retries* — instead of being consumed at boundaries where the bridge is down. EditorPrefs was also per-user machine-global (two concurrently open editors could consume/delete each other's pending resume) and survived crashes (a stale flag could resurrect the bridge on the next launch without opt-in). A once-per-session migration deletes the legacy key.
- **`TransportManager.StartAsync` serialized per mode**: concurrent starts coalesce onto one in-flight attempt. Manual Connect, reload-resume, and auto-start all flow through it, and `WebSocketTransportClient.StartAsync` tears down a live connection first — previously two interleaved starts could bounce a just-established session.
- **Connect-pending marker**: a reload that kills the in-flight connect phase no longer strands auto-start; the next domain load finishes *connect-only* (never re-spawns — `StartLocalHttpServer` stops a still-booting server first). `IServerManagementService.HasManagedServerLaunchHandle` answers the "can we watch the launch process die?" question live; without a handle the wait polls to the 5-minute hard cap instead of fail-fasting.
- **Busy gate** uses `EditorStateCache.GetActualIsCompiling` (now `internal`, reflection bound once as a delegate): raw `EditorApplication.isCompiling` stays true for a whole play session under Recompile-After-Finished-Playing (#549), which would otherwise block resume until play exits.
- User-intent cancel paths (Connect, End Session, transport switch, orphan cleanup) go through named seams (`CancelPendingResume` / `CancelPendingReconnect`) that also abort an in-flight retry loop.

## Intended behavior changes

- The HTTP bridge no longer auto-resumes across an editor *restart* unless Auto-Start on Load is enabled (the old EditorPrefs flag could trigger an unrequested launch-time resume after a crash).
- An invalid resume no longer replays a ~49s retry loop at every later reload: exhaustion erases the flag (matches the stdio sibling).
- A reload-interrupted auto-start now recovers (connect-only) instead of silently dying.

## Verification

- Compile-only matrix via `tools/check-unity-versions.sh` (2021.3.45f2 floor passes locally; CI covers the rest — the floor caught that `CompilationPipeline.isCompiling` needs reflection there).
- 23 EditMode tests added (`HttpBridgeReloadHandlerTests`, `HttpAutoStartHandlerTests`, `TransportManagerTests`), including a scenario test replaying the exact multi-pass-compile sequence from the issue, and a pin on the `StartAsync` coalescing contract. Full EditMode suite: 1002 passed / 0 failed locally.
- Note: the seam tests cover the decision/flag logic; the "delegate re-arms across a real domain reload" wiring can't be unit-tested on UTF 1.1 — verified manually. A repro pass on Windows per the issue's setup would be a welcome review step.

## Relationship to other open PRs (checked before filing)

- No open PR addresses #1229 (no cross-references on the issue).
- **#1121** touches `HttpAutoStartHandler` for a different claimed race (EditorPrefs unavailable during `[InitializeOnLoad]`) — it does not fix either root cause here, and it removes the once-per-session latch (a manually stopped bridge would resurrect after every reload). If that EditorPrefs issue is real (we could not corroborate it — the #1229 repro itself shows EditorPrefs working in the cctor), it can layer on top of this change.
- **#1073** (LAN transport) edits the URL-policy lines inside `AutoStartAsync` — trivial textual conflict with this PR, easy rebase for whichever lands second.

## Known follow-ups (deliberately out of scope)

- **stdio sibling has the same defect class**: `StdioBridgeReloadHandler.cs:65` (flag deleted at a bridge-down boundary — primary) and `:133` (`delayCall` deferral — secondary). Kept out to respect #688's centralized stdio teardown sequencing; same pattern applies.
- Remaining stdio copies of the `isCompiling` reflection probe can now consolidate onto `EditorStateCache.GetActualIsCompiling`.
- The manual Start Server path (`McpConnectionSection.TryAutoStartSessionAsync`) lacks the same reload survival; candidate for the same connect-pending pattern.

## Summary by CodeRabbit

* **New Features**
  * Added reload-safe handling for HTTP auto-start, reconnect resumption, and reload-resume tracking.
  * Improved transport startup so repeated start requests for the same transport share the same in-flight attempt.

* **Bug Fixes**
  * Prevented stale resume/reconnect flags from persisting across editor reloads and user transport/session changes (migrated resume state to session-level storage).
  * More accurate “is compiling” detection during Play Mode and compilation pipeline activity.
  * Exposed managed server launch-handle availability for clearer state reporting.

* **Tests**
  * Added edit-mode coverage for auto-start/reconnect decisions, reload-resume behavior, and concurrent transport start coalescing.

### DIFF
--- a/MCPForUnity/Editor/Constants/EditorPrefKeys.cs
+++ b/MCPForUnity/Editor/Constants/EditorPrefKeys.cs
@@ -17,7 +17,6 @@ namespace MCPForUnity.Editor.Constants
         internal const string DebugLogs = "MCPForUnity.DebugLogs";
         internal const string ValidationLevel = "MCPForUnity.ValidationLevel";
         internal const string UnitySocketPort = "MCPForUnity.UnitySocketPort";
-        internal const string ResumeHttpAfterReload = "MCPForUnity.ResumeHttpAfterReload";
         internal const string ResumeStdioAfterReload = "MCPForUnity.ResumeStdioAfterReload";
 
         internal const string UvxPathOverride = "MCPForUnity.UvxPath";
--- a/MCPForUnity/Editor/Services/EditorStateCache.cs
+++ b/MCPForUnity/Editor/Services/EditorStateCache.cs
@@ -529,12 +529,32 @@ namespace MCPForUnity.Editor.Services
             }
         }
 
+        // CompilationPipeline.isCompiling is not available on the 2021.3 package floor, hence
+        // reflection; bound once as a delegate so per-frame callers avoid the reflection invoke.
+        private static readonly Func<bool> PipelineIsCompiling = CreatePipelineIsCompilingDelegate();
+
+        private static Func<bool> CreatePipelineIsCompilingDelegate()
+        {
+            try
+            {
+                var prop = Type.GetType("UnityEditor.Compilation.CompilationPipeline, UnityEditor")
+                    ?.GetProperty("isCompiling", BindingFlags.Public | BindingFlags.Static);
+                return prop?.GetMethod != null
+                    ? (Func<bool>)prop.GetMethod.CreateDelegate(typeof(Func<bool>))
+                    : null;
+            }
+            catch
+            {
+                return null;
+            }
+        }
+
         /// <summary>
         /// Returns the actual compilation state, working around a known Unity quirk where
         /// EditorApplication.isCompiling can return false positives in Play mode.
         /// See: https://github.com/CoplayDev/unity-mcp/issues/549
         /// </summary>
-        private static bool GetActualIsCompiling()
+        internal static bool GetActualIsCompiling()
         {
             // If EditorApplication.isCompiling is false, Unity is definitely not compiling
             if (!EditorApplication.isCompiling)
@@ -543,18 +563,10 @@ namespace MCPForUnity.Editor.Services
             }
 
             // In Play mode, EditorApplication.isCompiling can have false positives.
-            // Double-check with CompilationPipeline.isCompiling via reflection.
-            if (EditorApplication.isPlaying)
+            // Double-check with CompilationPipeline.isCompiling.
+            if (EditorApplication.isPlaying && PipelineIsCompiling != null)
             {
-                try
-                {
-                    Type pipeline = Type.GetType("UnityEditor.Compilation.CompilationPipeline, UnityEditor");
-                    var prop = pipeline?.GetProperty("isCompiling", BindingFlags.Public | BindingFlags.Static);
-                    if (prop != null)
-                    {
-                        return (bool)prop.GetValue(null);
-                    }
-                }
+                try { return PipelineIsCompiling(); }
                 catch
                 {
                     // If reflection fails, fall back to EditorApplication.isCompiling
--- a/MCPForUnity/Editor/Services/EditorStateCache.cs
+++ b/MCPForUnity/Editor/Services/EditorStateCache.cs
@@ -259,6 +259,12 @@ namespace MCPForUnity.Editor.Services
                 EditorApplication.update += OnUpdate;
                 EditorApplication.playModeStateChanged += _ => ForceUpdate("playmode");
 
+                // Tracks whether an assembly compilation is actually running, for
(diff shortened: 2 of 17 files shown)


Scriptwonder:
@coderabbitai review

PR #1236, opened 2026-07-04
fix(server): stop silent cross-instance routing when multiple editors are connected (#1023)

Scriptwonder:
Addresses the dangerous half of #1023 (agents writing into the wrong project when 2+ editors are open).

## Problem

Two independent routing hazards on the Python server:

1. **stdio silently guesses the target editor.** With multiple editors connected and no instance selected, `UnityConnectionPool._resolve_instance_id` routes to the "most recently active" instance by heartbeat — nondeterministic per call, so consecutive tool calls can land in *different projects*. The reporter had editor settings overwritten, submodules force-deleted, and source files overwritten in the wrong project. HTTP was already hardened to raise `InstanceSelectionRequiredError` on ambiguity; stdio (the default transport) never got the same guard.
2. **Concurrent local clients share one active-instance slot.** `UnityInstanceMiddleware.get_session_key` falls back to a shared `"global"`/`client_id` key, so two clients on one server stomp each other's `set_active_instance` selection.

## Fix

- **stdio**: when no instance identifier is given, no default is pinned, and more than one instance is connected, raise a structured instance-selection-required error listing the available `Name@hash` ids (mirroring the HTTP hub's semantics) instead of silently picking by heartbeat. Single-instance auto-routing is unchanged.
- **middleware**: active-instance persistence is delegated to FastMCP session state, falling back to `client_id | user:{user_id} | "global"` only where session identity genuinely doesn't exist, so concurrent clients keep isolated selections.

## Verification

`cd Server && uv run pytest tests/ -q` → **1305 passed, 3 skipped** on this branch rebased onto current beta, including new `test_stdio_instance_resolution.py` (multi-instance refusal matrix) and reworked instance-routing/characterization tests.

## Relationship to other open work

- **#981** (explicit per-request routing) remains the right structural end-state; this PR is the minimal guard that removes the destructive-write hazard now, and is compatible with it.
- **#1194** (auto-select instance by launch directory) is a UX complement — with this guard in place, its auto-selection would remove most of the new "please select an instance" friction for multi-editor users.
- `UNITY_MCP_DEFAULT_INSTANCE` / `set_active_instance` remain the explicit opt-ins for scripted multi-instance workflows.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved active Unity instance tracking using per-session state for more reliable instance handling across requests.
  * When multiple instances exist and no default is set, instance resolution now requires an explicit selection (no more “most recent” guessing).
  * Instance-selection errors now include available instance IDs and return a `select_instance` hint through the transport; stdio bridge readiness and connection health checks were hardened (debounced verification and non-optimistic readiness).
* **Tests**
  * Updated/added integration, routing, stdio resolution, and transport characterization tests to match the new per-session behavior and error/hint details.

### DIFF
--- a/MCPForUnity/Editor/Services/Transport/Transports/StdioTransportClient.cs
+++ b/MCPForUnity/Editor/Services/Transport/Transports/StdioTransportClient.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Threading.Tasks;
 using MCPForUnity.Editor.Helpers;
+using UnityEditor;
 
 namespace MCPForUnity.Editor.Services.Transport.Transports
 {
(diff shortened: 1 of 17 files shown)


Scriptwonder:
@coderabbitai review

PR #1237, opened 2026-07-04
fix: stop tearing down healthy HTTP sessions on transient probe misses (#1207)

Scriptwonder:
Addresses the dominant churn source in #1207, following the fix order @dsarno proposed from the evidence bundle.

## Problem

Two amplifiers turned ordinary reload/test boundaries into `no_unity_session` wedges:

1. **Hair-trigger orphan detector.** `McpConnectionSection.UpdateConnectionStatus` ends an active session the moment one cached reachability reading is false — and that reading is a lone TCP connect with a **50 ms** wait, cached 0.75 s. On a machine busy with a test run or domain reload the probe times out against a perfectly healthy server; the detector then force-stops the bridge (13 teardowns / 147 socket closures in one session in the #1207 evidence bundle), feeding the reconnect churn that outruns the server's session-resolve window.
2. **Neutered escape hatch.** `UNITY_MCP_SESSION_RESOLVE_MAX_WAIT_S` is documented as configurable but was clamped to a ceiling equal to its default (20 s), so users whose reloads legitimately exceed 20 s could not actually extend the window.

## Fix

- Orphan teardown now requires **3 consecutive failed polls** at the 0.75 s cadence (~2.25 s of continuous unreachability); readings taken while the editor is compiling/importing don't count, and detection is skipped while busy. The decision is an `internal static` predicate with tests.
- Probe connect wait raised 50 ms → 250 ms, implemented as an **overall budget shared across candidate hosts** so the worst-case main-thread wait cannot multiply (a refused connect still fails in ~1 ms; the budget only bites when packets are dropped or the stack is starved).
- `UNITY_MCP_SESSION_RESOLVE_MAX_WAIT_S` ceiling raised to 120 s (default unchanged at 20 s); `UNITY_MCP_SESSION_READY_WAIT_SECONDS` likewise (default 6 s). Both now share one bounded env-read helper with tests.

## What this deliberately does not do

- No keepalive/reload-awareness changes in `WebSocketTransportClient` — #1234's resume rework already owns reload boundaries, and per the maintainer analysis the detector debounce removes the dominant churn source. If churn persists after both land, that's the next lever.
- No change to ping/pong heartbeat semantics from #656.

## Verification

- `cd Server && uv run pytest tests/ -q` → **1313 passed, 3 skipped** (5 new env-clamp tests).
- 7 new EditMode tests covering the teardown predicate matrix (threshold, below-threshold, busy-editor deferral, toggle-in-progress, non-HTTP-local, not-running).
- Compile-only matrix via `tools/check-unity-versions.sh` (2021.3 floor passes locally).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved local server reachability checks with more tolerant timing and a single overall timeout budget.
  * Reduced unexpected teardown of HTTP Local sessions by requiring multiple consecutive “server down” polls and deferring when the editor is busy or a connection toggle is underway.
  * Hardened wait-time configuration from environment variables, safely clamping values up to 120 seconds and falling back on invalid inputs.
* **Tests**
  * Added automated coverage for the wait-time configuration handling and orphaned-session detection logic.

### DIFF
--- a/MCPForUnity/Editor/Services/ServerManagementService.cs
+++ b/MCPForUnity/Editor/Services/ServerManagementService.cs
@@ -492,7 +492,10 @@ namespace MCPForUnity.Editor.Services
                     return false;
                 }
 
-                return TryConnectToLocalPort(uri.Host, uri.Port, timeoutMs: 50);
+                // 250ms, not 50ms: on a machine busy with test runs or domain reloads a 50ms
+                // connect wait produces false "server gone" readings that tore down healthy
+                // sessions via the orphaned-session detector (#1207).
+                return TryConnectToLocalPort(uri.Host, uri.Port, timeoutMs: 250);
             }
             catch
             {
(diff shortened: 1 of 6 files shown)


Scriptwonder:
@coderabbitai review

PR #1266, opened 2026-07-13
Drop a stale active-instance pin instead of failing forever

lgarczyn:
A pin outlives the editor it names.
While it stayed pinned it also suppressed auto-select. Every later call then failed with no_unity_session. Neither waiting nor relaunching the editor recovered. The editor re-registers under its own name, not the name the pin holds. The pin is now dropped once another instance is registered. An empty registry means a domain reload is in flight, so the pin is kept.

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Documentation update
- [ ] Test update

- Unity version(s) tested:
- Package source used (`#beta`, `#main`, tag, branch, or `file:`):
- Resolved commit hash from `Packages/packages-lock.json` (if using a Git package URL):

- [ ] Unity EditMode tests
- [ ] Unity PlayMode tests
- [ ] Package import/compile check

## Documentation Updates

## Summary by CodeRabbit

* **Bug Fixes**
  * Detects when a selected Unity instance is no longer available and clears the stale selection.
  * Reports a clear error instead of silently switching to another project.
  * Preserves selections during temporary periods when no instances are registered or while an instance is reconnecting.

### DIFF
--- a/Server/src/transport/unity_instance_middleware.py
+++ b/Server/src/transport/unity_instance_middleware.py
@@ -212,6 +212,35 @@ class UnityInstanceMiddleware(Middleware):
             "Read mcpforunity://instances for current sessions."
         )
 
+    async def _drop_stale_pin(self, ctx, active_instance: str) -> str | None:
+        """
+        Discard a pinned instance that no plugin is registered under.
+
+        A pin outlives the editor it names. Left in place it also suppresses
+        auto-select, so every later call fails with no_unity_session and neither
+        waiting nor relaunching the editor recovers: the editor re-registers
+        under its own name, which is not the name the pin holds.
+
+        An empty registry means a domain reload is in flight and the pin is
+        still good, so only drop it while some other instance is registered.
+        """
+        from transport.unity_transport import _is_http_transport
+        if not (_is_http_transport() and PluginHub.is_configured()):
+            return active_instance
+
+        instances = await self._discover_instances(ctx)
+        ids = [inst.id for inst in instances if getattr(inst, "id", None)]
+        if not ids or active_instance in ids:
+            return active_instance
+
+        _diag.warning(
+            "Active instance %s is no longer registered; dropping the stale pin. Registered: %s",
+            active_instance,
+            ", ".join(ids),
+        )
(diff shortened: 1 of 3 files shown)


Scriptwonder:
Hi there, I think this might need to be rebased as well first. Thanks for the PR and let me know if you can resolve these comments!

Two changes are needed, and website/docs/architecture/instance-routing.md:181-183 specifies the shape: "Clearing the pin automatically is not the fix. Clearing it and then selecting the one remaining Editor is #1023's exact harm arriving through a different door. The acceptable shape is to clear only after the reconnect wait has expired for that specific hash, and never to retarget silently." 
So first, gate the drop on the per-hash reconnect wait that _resolve_session_id already performs (UNITY_MCP_SESSION_RESOLVE_MAX_WAIT_S, default 20s — plugin_hub.py:886-887); _drop_stale_pin currently runs upstream of that poll, so a routine domain reload trips it before the wait ever gets a chance. Second, after dropping, raise an error naming the departed instance instead of returning None and falling through to _maybe_autoselect_instance — as written, a call pinned to project A lands in project B, returns success, and re-pins B.

lgarczyn:
> Hi there, I think this might need to be rebased as well first. Thanks for the PR and let me know if you can resolve these comments!
> 
> Two changes are needed, and website/docs/architecture/instance-routing.md:181-183 specifies the shape: "Clearing the pin automatically is not the fix. Clearing it and then selecting the one remaining Editor is #1023's exact harm arriving through a different door. The acceptable shape is to clear only after the reconnect wait has expired for that specific hash, and never to retarget silently." So first, gate the drop on the per-hash reconnect wait that _resolve_session_id already performs (UNITY_MCP_SESSION_RESOLVE_MAX_WAIT_S, default 20s — plugin_hub.py:886-887); _drop_stale_pin currently runs upstream of that poll, so a routine domain reload trips it before the wait ever gets a chance. Second, after dropping, raise an error naming the departed instance instead of returning None and falling through to _maybe_autoselect_instance — as written, a call pinned to project A lands in project B, returns success, and re-pins B.

All done

Issue #1138, opened 2026-05-19
Public UnityEngine.Object JsonConverter can disrupt existing Newtonsoft.Json serialization


n-starshov:
## Summary
`MCPForUnity.Runtime.Serialization.UnityEngineObjectConverter` is a public `JsonConverter<UnityEngine.Object>` with a default constructor.
This makes it discoverable by packages/tools that scan loaded assemblies for Newtonsoft `JsonConverter` implementations and add them to global `JsonConvert.DefaultSettings`.
As a result, installing Unity MCP can unintentionally change unrelated project serialization behavior.

## Problem
The converter serializes Unity assets as asset paths:
``` csharp
if (UnityEditor.AssetDatabase.Contains(value))
{
    string path = UnityEditor.AssetDatabase.GetAssetPath(value);
    writer.WriteValue(path);
}
```

That behavior is useful for MCP-internal serialization, but it is risky when the converter becomes globally discoverable.
Any unrelated code that serializes an object containing a ScriptableObject, Material, Texture, prefab reference, or other UnityEngine.Object may start receiving asset paths instead of normal object JSON.

### Minimal Example

``` csharp
public class SomeConfig : ScriptableObject
{
    public List<Row> Rows = new();
}
public struct Container
{
    public SomeConfig Config;
}
var json = JsonConvert.SerializeObject(container);
```

Expected by unrelated project code:
``` json
{
  "Config": {
    "Rows": []
  }
}
```

Possible result after global converter auto-discovery:

``` json
{
  "Config": "Assets/Path/To/SomeConfig.asset"
}
```

### Why This Is Surprising

The affected code does not call MCP APIs or MCP serializer settings directly. The behavior can change just because the MCP package is installed and its converter type is visible to external converter discovery.
One common trigger is jillejr.newtonsoft.json-for-unity.converters, which can scan loaded assemblies for public Newtonsoft converters and inject them into JsonConvert.DefaultSettings.
Suggested Fixes
Possible options:
1. Make UnityEngineObjectConverter internal if it is only intended for MCP internals.
2. Remove the public parameterless constructor so generic converter scanners do not instantiate it automatically.
3. Move MCP-specific converters into an internal/editor-only assembly not intended for global discovery.
4. Avoid exposing broad converters such as JsonConverter<UnityEngine.Object> publicly.
5. Keep the converter public but require explicit MCP serializer settings to use it.

## Environment
- Unity 6000.0.60f1
- com.coplaydev.unity-mcp
[...]

PR #1274, opened 2026-07-16
fix: guard TransformHandle against reflection serialization (destroyed-object NRE)

agudmund:
## Why

`GameObjectSerializer` reflects over component properties and hands the values to Json.NET. On Unity 6, a component property returning `UnityEngine.TransformHandle` makes the serializer enumerate the handle's direct children (`TransformHandle.DirectChildrenEnumerator`). Two problems fall out of that:

- **A play-mode race:** the enumeration throws `NullReferenceException` from `TransformHandle.AssertHandleIsValid` the moment the underlying object is destroyed. Reading the `gameobject/{id}/components` resource while a game script `Destroy()`s objects (a spawner destroying and respawning pickups makes it reliable) logs the `[GameObjectSerializer] Unexpected error serializing value of type UnityEngine.TransformHandle` warning on every read that races a destroy.
- **Even on live objects,** serializing a handle walks child hierarchies of further handles rather than producing a meaningful value.

## What

A write-only `TransformHandleConverter` that emits `null` for the type. It matches by full type name (`UnityEngine.TransformHandle`) so the package still compiles on every supported Unity version (floor 2021.3, where the type does not exist) without version defines, per the `Unity*Compat` guidance in CONTRIBUTING. It is `internal` for the same third-party converter-scanner reason as `UnityEngineObjectConverter` (#1138), and registered in `GameObjectSerializer`'s output settings.

EditMode tests resolve the type by reflection and self-ignore on Unity versions without it; the key regression test obtains a handle, destroys its object, and asserts serialization no longer throws.

Observed in the field on Unity 6000.4.10f1 while polling component resources during play mode.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Unity 6 compatibility when serializing `TransformHandle` values.
  * Prevented serialization errors for handles linked to destroyed objects.
  * Transform handles are safely represented as `null` in serialized output.

* **Tests**
  * Added coverage for supported, unsupported, live, and destroyed `TransformHandle` scenarios.

### DIFF
--- a/MCPForUnity/Editor/Helpers/GameObjectSerializer.cs
+++ b/MCPForUnity/Editor/Helpers/GameObjectSerializer.cs
@@ -714,6 +714,7 @@ namespace MCPForUnity.Editor.Helpers
                 new RectConverter(),
                 new BoundsConverter(),
                 new Matrix4x4Converter(), // Fix #478: Safe Matrix4x4 serialization for Cinemachine
+                new TransformHandleConverter(), // Unity 6 TransformHandle: enumerating a destroyed handle's children throws NRE
                 new UnityEngineObjectConverter() // Handles serialization of references
             },
             ReferenceLoopHandling = ReferenceLoopHandling.Ignore,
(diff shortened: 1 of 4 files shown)


Scriptwonder:
Hi thanks for the PR. Can you double make sure its NRE and not MRE, MissingReferenceException? Ive repro the other term rather than NRE. Just want to make sure the comments are written clearly and we also handle both cases.

PR #1300, opened 2026-07-29
fix: deduplicate CodeDom assembly references

kpkhxlgy0:
## Description

Fix CodeDom dynamic compilation failures when Unity has multiple loaded assemblies with the same simple name. The observed failure involved `System.Security` 2.0.0.0 and 4.0.0.0: excluding `System.Security` entirely avoided the duplicate-reference error but made types defined only in that assembly unavailable.

## Type of Change

- [ ] Documentation update
- [x] Test update

## Changes Made

- Preserve the existing netstandard facade exclusions used by the CodeDom backend.
- Deduplicate remaining CodeDom references by managed assembly simple name.
- Prefer the exact assembly identity referenced by the most loaded assemblies, then the higher version, then a stable path tie-break.
- Preserve `System.Security` instead of removing it from CodeDom references.
- Add portable EditMode regression coverage using temporary generated assemblies, without project-specific DLL paths.

- Unity version(s) tested: Unity 2022.3.23f1 on Windows
- Package source used (`#beta`, `#main`, tag, branch, or `file:`): local file checkout based on `CoplayDev/unity-mcp#beta` at `fc70dda7`
- Resolved commit hash from `Packages/packages-lock.json` (if using a Git package URL): not applicable (`file:` checkout)

- [ ] Unity EditMode tests
- [ ] Unity PlayMode tests

Validation in the active Unity 2022.3 project:

- Package scripts compiled with no `ExecuteCode` or `error CS` Console entries.
- CodeDom selected `System.Security, Version=4.0.0.0` from simultaneous 2.0.0.0 and 4.0.0.0 inputs.
- CodeDom compiled and resolved `System.Security.Cryptography.ProtectedData` from `System.Security, Version=4.0.0.0`.
- Equivalent temporary-assembly checks preserved a file named `System.Security.dll` with `netstandard` present and selected referenced version 1.0.0.0 over unreferenced version 2.0.0.0.

## Documentation Updates

No tool or resource surface changed.

## Related Issues

None.

## Additional Notes

The two Unity EditMode regression tests were added to `ExecuteCodeTests.cs`, but the standalone `UnityMCPTests` project was not opened or run locally. The active host project was used for package compilation and equivalent runtime checks.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CodeDom compilation reliability when projects include multiple versions of the same assembly.
  * Preserved required security assemblies during compilation with netstandard references.
  * Added smarter assembly selection to favor versions referenced by the running Unity environment.
  * Improved repeated compilation performance by reusing filtered assembly paths until the Unity domain reloads.

* **Tests**
  * Added regression coverage for assembly filtering, duplicate assembly resolution, security assemblies, and cache reuse.

### DIFF
--- a/MCPForUnity/Editor/Tools/ExecuteCode.cs
+++ b/MCPForUnity/Editor/Tools/ExecuteCode.cs
@@ -30,11 +30,13 @@ namespace MCPForUnity.Editor.Tools
 
         private static readonly List<HistoryEntry> _history = new List<HistoryEntry>();
         private static string[] _cachedAssemblyPaths;
+        private static string[] _cachedCodeDomAssemblyPaths;
 
         [UnityEditor.InitializeOnLoadMethod]
         private static void OnDomainReload()
         {
             _cachedAssemblyPaths = null;
+            _cachedCodeDomAssemblyPaths = null;
             RoslynCompiler.ResetCache();
         }
 
@@ -342,6 +344,10 @@ namespace MCPForUnity.Editor.Tools
 
         internal static string[] FilterAssemblyPathsForCodeDom(string[] allPaths)
         {
+            var useCache = ReferenceEquals(allPaths, _cachedAssemblyPaths);
+            if (useCache && _cachedCodeDomAssemblyPaths != null)
+                return _cachedCodeDomAssemblyPaths;
+
             var hasNetstandard = allPaths.Any(p =>
                 string.Equals(Path.GetFileNameWithoutExtension(p), "netstandard", StringComparison.OrdinalIgnoreCase));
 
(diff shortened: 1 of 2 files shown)


Scriptwonder:
Hi, thanks for the PR and identify the issue. One thing before merge: is it possible to also cache the CodeDomPaths instead of calling getAseemblyName on every path on every execute_code call? Just thinking on ways to optimize this, lmk what you think.

Scriptwonder:
Nice catch on the root cause, and the version-preference tests are the right shape.

One thing before merge: `DeduplicateAssemblyPathsForCodeDom` calls `AssemblyName.GetAssemblyName()` on every path — 200–500 in a typical Editor, each one opening the DLL to read its manifest — on every CodeDom `execute_code` call. Since assemblies only change across a domain reload, this can just be memoized next to `_cachedAssemblyPaths` and cleared in the existing `OnDomainReload()` (`ExecuteCode.cs:32-39`), which makes it a once-per-reload cost instead of once-per-call.

Guard the memo with `ReferenceEquals(allPaths, _cachedAssemblyPaths)` so the two new tests keep hitting the uncached path with their own arrays.

Also, the red CI isn't yours: `actions/checkout` now refuses fork checkouts under `pull_request_target`, and we haven't set `allow-unsafe-pr-checkout` yet. Tracking that separately — nothing for you to chase.

PR #1285, opened 2026-07-28
Scope HTTP endpoint and server ownership per project

liuzqk:
## Description

Fix machine-global `EditorPrefs` cross-contamination that can make one Unity project inherit another project's local HTTP endpoint or server ownership metadata.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [x] Test update

## Changes Made

- Scope local endpoint, transport selection, PID, port, PID-file path, and instance token by project hash.
- Let `UNITY_MCP_HTTP_URL` override only the current Unity process and reject non-local or unsupported schemes.
- Migrate the legacy endpoint/transport bundle atomically to at most one project under a cross-process mutex.
- Keep known global preferences global and use one storage-key resolver for all scoped accessors and diagnostics.
- Display/search the resolved project key in the EditorPrefs diagnostics window.
- Preserve pre-existing scoped endpoint/PID state in test fixtures and disable migration side effects during snapshot and teardown refreshes.

## Compatibility / Package Source

- Unity version(s) tested: 2022.3.62f3
- Package source used (`#beta`, `#main`, tag, branch, or `file:`): branch derived directly from official `beta` (`bd72241ab91c664176091789c9408d676783abec`); local `file:` source for validation only
- Resolved commit hash from `Packages/packages-lock.json` (if using a Git package URL): N/A

Single-project defaults remain unchanged. Configuration remains machine-local and is not written into the Unity project.

## Testing/Screenshots/Recordings

- [ ] Python tests (`cd Server && uv run pytest tests/ -v`)
- [x] Unity EditMode tests
- [ ] Unity PlayMode tests
- [x] Package import/compile check
- [ ] Not applicable (explain why in Additional Notes)

Unity 2022.3.62f3 targeted suite: 155 total, 154 passed, 0 failed, 1 intentional stress-only skip. Two isolated Unity processes also validated independent endpoints and project-scoped PID tracking.

## Documentation Updates

- [ ] I have added/removed/modified tools or resources
- [ ] If yes, I have updated all documentation files using:
  - [ ] The LLM prompt at `tools/UPDATE_DOCS_PROMPT.md` (recommended)
  - [ ] Manual review of the generated changes

## Related Issues

Relates to #1164.

## Additional Notes

[...]

PR #1208, opened 2026-06-16
Feat/optional physics screencapture modules

ArkTarusov:
## Description
9.7.1 made MCP for Unity hard-depend on three built-in modules (Physics, Physics 2D,
Screen Capture): `UnityPhysicsCompat.cs` uses `typeof(Physics)`/`typeof(Physics2D)` and
several editor tools reference Physics/Physics2D/ScreenCapture types directly, so the
package fails to compile when any of those modules is disabled (#1172). They were also
added as hard `package.json` dependencies, which force-pulls them into every consuming
project — undesirable for an editor-focused package. This PR makes all three modules
optional via `versionDefines` (the approach proposed in the issue) and removes the forced
dependencies.

## Type of Change
- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Test update

## Changes Made
- Added `versionDefines` to `MCPForUnity.Editor.asmdef` and `MCPForUnity.Runtime.asmdef`
  (mirroring the existing `UNITY_VFX_GRAPH`): `MCP_HAS_PHYSICS`, `MCP_HAS_PHYSICS_2D`,
  `MCP_HAS_SCREEN_CAPTURE`, resolved from `com.unity.modules.physics`,
  `com.unity.modules.physics2d`, `com.unity.modules.screencapture`.
- `Runtime/Helpers/UnityPhysicsCompat.cs`: resolve `Physics`/`Physics2D` via reflection
  (`Type.GetType(...)`) instead of `typeof`, so the shim compiles with the modules absent
  (no `#if`; consistent with the file's existing reflection design).
- `Runtime/Helpers/ScreenshotUtility.cs`: guarded all `ScreenCapture` call sites behind
  `MCP_HAS_SCREEN_CAPTURE`, falling back to camera-based capture when absent.
- Guarded all Physics/Physics2D usage in the editor tools behind the defines:
  `Editor/Tools/Physics/*` (incl. `ManagePhysics` returning a clear "module not installed"
  message), plus incidental usages in `ManageScene`, `ManageAsset`, `Helpers/ComponentOps`,
  `Helpers/GameObjectSerializer`, `Tools/GameObjects/GameObjectComponentHelpers`.
  3D and 2D physics are handled independently.
- Removed `com.unity.modules.physics`, `com.unity.modules.physics2d` and
[...]

PR #1206, opened 2026-06-15
Extend animator controller editing capabilities

Thaina:
## Description
adding `get_state_properties` and `set_state_properties` to the mcp managed animation controller so that AI can modified the node graph position in addition to create and set/remove its transition (RemoveTransition added too)

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Test update

## Compatibility / Package Source

- Unity version(s) tested: 6000.3.13f
- Package source used : file:
- Resolved commit hash from `Packages/packages-lock.json`: 4d8cf5dedfb518221ade4741136a2e2b62cc7218

## Testing/Screenshots/Recordings

- [x] Python tests (`cd Server && uv run pytest tests/ -v`)
- [x] Unity EditMode tests
- [x] Unity PlayMode tests
- [x] Package import/compile check
- [ ] Not applicable (explain why in Additional Notes)

## Documentation Updates

- [x] I have added/removed/modified tools or resources
- [ ] If yes, I have updated all documentation files using:
  - [ ] The LLM prompt at `tools/UPDATE_DOCS_PROMPT.md` (recommended)
  - [x] Manual review of the generated changes

## Summary by CodeRabbit

* **New Features**
  * Added tools to inspect and update animation state properties, including position, speed, and motion assignments.
  * Added transition removal and expanded transition configuration options.
  * Added blend-tree inspection and editing, including child clip and nested tree support.
  * Added animation clip root-motion analysis with sampled paths and movement metrics.
  * Improved asset references with lossless instance identifiers across supported Unity versions.
* **Bug Fixes**
  * Improved validation and error reporting for animation controller, blend-tree, and clip operations.
* **Tests**
  * Added coverage for new animation actions and parameter forwarding.

PR #1123, opened 2026-05-13
feat(prefabs): add refresh flag to save_prefab_stage to skip AssetDatabase.Refresh()

sssooonnnggg:
Allows callers to opt out of the trailing SaveAssets()+Refresh() (often ~1s on large projects) and trigger refresh_unity themselves when needed.

## Summary by CodeRabbit

* **New Features**
  * Save and close prefab operations now accept an optional refresh flag to control whether the editor runs the post-save refresh step.
  * Close operation respects the refresh choice when saving before closing.

* **Documentation**
  * Usage note and performance tip added: opting out of automatic refresh avoids the trailing import delay; manually trigger refresh when needed.

[Review Change Stack](https://app.coderabbit.ai/change-stack/CoplayDev/unity-mcp/pull/1123)

PR #1323, opened 2026-08-12
feat: Unity 2020.3 LTS support (C#8 + netstandard2.0 + 2021.2 API guards)

RoyougiShiki:
## Summary

Enable **Unity 2020.3 LTS** support for MCP for Unity. The package currently requires 2021.3+; this PR lowers the floor to 2020.3 by removing C# 9 syntax and shimming .NET Core 2.1+ / Unity 2021.2+ APIs that 2020.3 lacks.

**All changes are behavior-neutral on 2021.3+**: newer Unity versions take the same code paths as before (guarded with `#if UNITY_2021_2_OR_NEWER`), and the 2020.3 branches are functionally equivalent implementations, not stubs.

Verified on Unity **2020.3.24f1**: full package compile with **0 errors / 0 warnings**, all 11 UXML files load, editor windows open, dropdown control + tool routing runtime checks pass. A dedicated test project is included (`TestProjects/Unity2020Compat`).

### 1. C# 9 → C# 8 (2020.3 ships the C# 8 compiler)
- Target-typed `new()` / `new(...)` → explicit types (~65 sites)
- `is not T x` pattern → `!(x is T x)` (20 sites)
- `is A or B` / property-pattern `or` combos → equivalent boolean expressions
- Switch-expression `or` arms → split arms; target-typed ternaries → explicit `IMcpResponse` casts

### 2. .NET API shims (netstandard2.0 vs 2.1)
- `string.Contains(char)`, `Contains(str, StringComparison)`, `string.Join(char, …)` → netstandard2.0 equivalents
- `Index/Range` slicing (`s[..^n]`) → `Substring`
- `Math.Clamp` → `Mathf.Clamp`, `Task.IsCompletedSuccessfully` → `TaskStatus.RanToCompletion`
- `Dictionary.Remove(k, out v)` → `TryGetValue` + `Remove`
- `ProcessStartInfo.ArgumentList` → new `AddArg()` extension (argument-quoting equivalent)
- `Path.GetRelativePath`, `Enum.TryParse(Type,…)` 4-arg, `Rfc2898DeriveBytes` 4-arg → 2020.3-compatible equivalents

### 3. Unity 2021.2+ APIs behind version guards (2020.3 branches are equivalents, not stubs)
- `NamedBuildTarget` → `BuildTargetGroup` (2020.3 has all the same PlayerSettings overloads)
- `PrefabStage`/`PrefabStageUtility` → `UnityEditor.Experimental.SceneManagement` (moved in 2021.2), `OpenPrefab` → `AssetDatabase.OpenAsset` + `GetCurrentPrefabStage`
- `PackageInfo.GetAllRegisteredPackages()` → new `RegisteredPackageInfo` helper: 2021.2+ wraps the native API; 2020.3 parses authoritative `Packages/packages-lock.json` (synchronous file IO — the alternative `Client.List` polling **deadlocks the editor**, see commit 3c690f0)
- `Client.AddAndRemove` → per-package `Add`/`Remove`; `StandaloneBuildSubtarget`/`CleanBuildCache` guarded

### 4. UI Toolkit
- `DropdownField` (2021.2+) → new `CompatDropdownField`: 2021.2+ inherits `DropdownField` unchanged; 2020.3 self-draws an equivalent popup (PopupField\<string\> keeps `choices` private) with UxmlFactory/UxmlTraits. 4 UXML files updated.

### 5. Documented 2020.3-only gaps (no equivalent API exists anywhere in 2020.3)
- `UIDocument` runtime UI tools return an explicit "requires Unity 2021.2+" error (tools stay registered)
- `LightingSettings.lightmapCompression`, `BuildOptions.CleanBuildCache`, standalone Server subtarget, `ProfilerCategory.FileIO/VirtualTexturing` → guarded/skipped on 2020.3

## Notes for maintainers
- Full patch inventory & sync strategy: `docs/UNITY_2020_3_COMPAT.md`
- Test project: `TestProjects/Unity2020Compat` (2020.3.24f1, `file:` link to `../../MCPForUnity`, one-click `verify_compile.cmd`)
- If you'd rather not commit to a 2020.3 floor, the C# 8 + .NET shim changes (sections 1–2) are independently valuable and mergeable on their own.

## Summary by CodeRabbit

* **New Features**
  * Added support for Unity 2020.3, including compatible package management, build settings, prefab stages, graphics, secure process arguments, and UI Toolkit controls.
  * Added cross-version dropdown controls for editor settings and configuration panels.

* **Bug Fixes**
  * Improved package detection and metadata handling across supported Unity versions.
  * Added graceful fallbacks for features unavailable in older Unity releases.

* **Documentation**
  * Added Unity 2020.3 compatibility guidance and a validation project.


RoyougiShiki:
Thanks for the thorough review! All actionable findings addressed in `e940e904` (11 files, 231+/31-):

**Fixed:**
- **ProcessArgumentListCompat** — rewrote quoting with the standard CommandLineToArgvW algorithm: backslashes doubled only before quotes/at end of argument, verbatim pass-through when no quoting needed.
- **EncryptedFileKeyStore** — 2020.3 now implements PBKDF2-HMAC-SHA256 manually (RFC 2898) since netstandard2.1 lacks the 4-arg `Rfc2898DeriveBytes`; key derivation is byte-identical with the 2021.2+ path, so existing ciphertext keeps validating.
- **CompatDropdownField.SetValueWithoutNotify** — no longer dispatches change events on 2020.3 (matches native 2021.2+ behavior).
- **MCPForUnityEditorWindow** — legacy (2020.3) package add/remove now serialized through a queue; next request starts only after the previous completes.
- **BuildSettingsHelper** — architecture mapping corrected: `0` = None (not x86_64); `x86_64` removed from valid inputs.
- **ManageBuild** — `subtarget: "server"` now returns an explicit error on <2021.2 instead of silently building player.
- **docs/UNITY_2020_3_COMPAT.md** — corrected to document `packages-lock.json` parsing (not `Client.List` polling).
- **Test project** — test-framework aligned to 1.1.31 across manifest/lock; `verify_compile.cmd` rewritten (portable: `%~dp0` + auto-detect Unity path or arg, exit codes, consistent log name, enforces `error CS` absence + clean batchmode exit); README command block fixed.

**Deferred with rationale (kept as-is for this PR):**
- Centralizing version guards into a `Unity*Compat.cs` shim (ManageBuild/BuildRunner/BuildTargetMapping): agreed it's the right long-term shape, but it's a cross-cutting refactor of upstream-owned files; keeping the `#if` guards local makes the 2020.3 delta easier to review and rebase. Happy to follow up with a shim refactor PR if maintainers prefer.
- `lightmapCompression` / `clean_build` silent skip on 2020.3: both are documented in the compat doc as APIs that don't exist before 2021.2; failing loudly would break existing tool callers on 2020.3. (ManageBuild's `server` subtarget is different — it would silently produce a *wrong* build, hence the explicit error there.)
- Docstring coverage warning: this PR is mechanical compatibility rewriting; docstrings were not added to unchanged logic. Can add if required by repo policy.

RoyougiShiki:
Second round addressed in `df74f8f0`:

- **verify_compile.cmd**: `%%~E` now strips the surrounding quotes when detecting the Unity executable (`%%E` produced `""C:\...\Unity.exe""`); Unity's exit code is now checked **before** accepting log markers, so a failed invocation can't report PASS against a stale log.
- **ManageBuild**: subtarget processing moved **before** the active-target short-circuit — a `server` request on Unity < 2021.2 now returns the explicit unsupported error even when the platform is already active (previously it silently returned "Already on this platform").
- **EncryptedFileKeyStore**: comment corrected to `netstandard2.0` (Unity 2020.3 API compatibility level), avoiding future misuse of 2.1-only APIs.

RoyougiShiki:
These four comments reference the pre-`df74f8f0` state — all four were already fixed in that commit:

- **verify_compile.cmd `%%E` quotes** → now `%%~E` (line 27)
- **verify_compile.cmd exit-code check** → Unity's `UNITY_STATUS` is checked before accepting log markers (line 48)
- **ManageBuild subtarget ordering** → subtarget processing now runs before the active-target short-circuit (line 235)
- **EncryptedFileKeyStore netstandard comment** → corrected to `netstandard2.0` (line 106)

Verified compiling on Unity 2020.3.24f1 (0 errors / 0 warnings). Ready for the next review round on `df74f8f0`.

RoyougiShiki:
Addressed in `775f15b1`: the `if exist` operand is now quoted (`if exist "%%~E"`), keeping the assignment unquoted — handles spaces in default Hub install paths (`C:\Program Files\...`). Verified on cmd.exe with a space-containing test path.

(Note: Blinter's `E024` on `%~E` is a false positive — `%%~E` is the standard for-variable quote-stripping syntax, not a `%1`-style parameter modifier.)

RoyougiShiki:
Round 4 addressed in `639279b4`:

- **Exit-code contract** — documented: `1` = verification failed (compile errors, Unity startup failure, missing log, unclean exit); `0`/`2` meanings retained.
- **Editor version check** — the chosen editor (arg or `UNITY_EDITOR`) is now checked for a `2020.3` path component with a clear warning when it isn't. Not a hard reject: passing a newer editor is a legitimate forward-compat check, but the warning makes it explicit that 2020.3-floor validation requires the 2020.3 editor.
- **Matrix note** — PASS output now points to `tools/check-unity-versions.sh` for the full supported-version matrix.

Skipped with rationale: in-script timeout/termination for the Unity process — this script is a local/CI helper where the caller already bounds the job (our runs use an outer timeout); adding `start /w` + `taskkill` bookkeeping inside the .cmd would add failure modes without real coverage. Happy to add it if a CI job needs to rely on the script alone.

Scriptwonder:
Hi sorry for the wait, but I would like sometime to review it. Does this huge PR gonna introduce regression for the new version? We are going to introduce a lot of deprecated/obsolete APIs into the repo for this case.

Scriptwonder:
Thanks for this — 96 files of version-guard work is a serious amount of effort, and I want to be straight with you about where it stands rather than leave it sitting silently.

**I'm not closing this.** But I can't merge or even properly review it in its current state, and I'd rather explain the blockers concretely so you can decide whether to push it forward.

### 1. It doesn't compile on the versions we ship today

`Compile MCPForUnity (win/osx/linux)` has been red since 2026-08-15 ([run 31568516031](https://github.com/CoplayDev/unity-mcp/actions/runs/31568516031)), failing identically on all three OS:

```
MCPForUnity/Editor/Tools/Graphics/SkyboxOps.cs(18,17): error CS0266:
Cannot implicitly convert type 'UnityEngine.Texture' to 'UnityEngine.Cubemap'.
```

`SkyboxOps.cs:14` narrows the property to `Cubemap`, but under `UNITY_2022_1_OR_NEWER` the getter returns `Texture`. So as it stands this breaks every user on 2022.1+ — the current user base — in order to add 2020.3.

Worth flagging why this wasn't obvious: the other checks on this PR look green, but `Test in editmode` and `e2e-bridge` **skip and still report success** when license secrets aren't in scope, which is the case for every fork PR. On this repo, `Compile MCPForUnity` and `Run Python Tests` are the only checks that mean anything from a fork. That's our problem, not yours, and I'm fixing it — but it's why the red one is the one to watch.

### 2. The version guards don't go where CLAUDE.md says they go

The PR adds **27 new `#if UNITY_*` directives** at call sites. Our policy (see the XML doc on `MCPForUnity/Runtime/Helpers/UnityCompatShims.cs`, which is the source of truth and ships inside the UPM package) is that a renamed/removed API gets a shim in `Unity*Compat.cs` and every caller routes through it, precisely so we don't have to audit N call sites per version bump. None of the new gates are registered in that catalog.

### 3. The half this exists for can't be verified

`tools/unity-versions.json` has no 2020.3 row, so neither CI nor `tools/check-unity-versions.sh` ever compiles the configuration this PR is for. Adding 2020.3 support without adding it to the matrix means the next unrelated PR silently breaks it.

### 4. What it interrupts

This is the part I'd weigh hardest. The branch overlaps files with **12 other open PRs**, so whatever order these land in, most of them will need a rebase:

| PR | overlapping files | |
|---|---|---|
| #1121 | 12 | `PortManager.cs`, `StdioBridgeHost.cs`, `UnityTypeResolver.cs`, … |
| #1208 | 4 | `GameObjectComponentHelpers.cs`, `ManageScene.cs`, `ManageUI.cs`, `package.json` |
| #1285 | 4 | `CodexConfigHelper.cs`, `HttpEndpointUtility.cs`, `ProjectIdentityUtility.cs`, … |
| #1327 | 3 | `CommandRegistry.cs`, `ManageScene.cs`, `ManageUI.cs` |
| #1206 | 2 | `ClipCreate.cs`, `ControllerCreate.cs` |
| #1035 | 2 | `BatchExecute.cs`, `UnityTypeConverters.cs` |
| #1282 #1342 #1347 #1123 #1274 #1287 | 1 each | |

There's also a standing constraint I'd be signing up for: targeting C# 8 means every future C# PR has to avoid range operators, `is not`, target-typed `new` and so on, and we have no lint gate that would catch a violation. That's a real ongoing tax and I want to be honest that I haven't decided to accept it.

### If you want to carry it forward

In rough order of what would change my answer:

1. Fix the `SkyboxOps.cs` CS0266 so the compile check goes green — nothing else can be assessed until it does.
2. Add a `2020.3.x` row to `tools/unity-versions.json` so the target is actually exercised.
3. Route the version-gated APIs through `Unity*Compat.cs` shims and register them in the `UnityCompatShims.cs` catalog, instead of `#if` at call sites.

If splitting helps, the guards that are just *correctness on versions we already support* would be a much easier PR to land on their own, separately from the floor change.

Either way — thank you for the work, and sorry it took this long to give you a real answer.

msanatan:
I'd also add to @Scriptwonder 's feedback, supporting many versions of Unity is hard. There are nuances across each one. For example, if this project were to focus on Unity 6, especially as 7 is backward compatible, the testing radius and coding standards would be much easier to manage.

2020 works in different ways that adds nontrivial maintenance overhead to support. Something to keep in mind!

PR #1330, opened 2026-08-16
docs: fix broken star history charts

FaintFlower:
The star history charts in the READMEs are currently broken because the GitHub stargazer API they depend on no longer allows unauthenticated access. This swaps them over to star-history.dera.page, which serves the same charts from an alternative data source that requires no API token and uses the same domain for both the image and the link.

## Summary by CodeRabbit

* **Documentation**
  * Updated the Star History chart and link to use the new service address.
  * Applied the update to both English and Chinese documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ See [CONTRIBUTING.md](CONTRIBUTING.md). Branch off `beta`, not `main`. The full 
 
 ## Star History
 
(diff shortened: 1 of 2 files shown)


Scriptwonder:
Eh thanks for this! I've been longing for this update and glad it is done by the community. Will merge it right away. Much appreciated!

PR #1342, opened 2026-08-23
fix(read_console): stop the Console window's filters from hiding entries

### LINKED ISSUE #1239: [Bug] read_console silently returns 0 entries when Console window severity toggles are off (filter inheritance)
## Description

`read_console` silently returns `success: true` with **zero entries** whenever the Unity Console window's severity toggle buttons (the Log / Warning / Error counters in the Console toolbar) are switched off — even for freshly logged messages, and even though the logs are plainly in `Editor.log`. An agent asking for `types: ["log"]` has no way to distinguish "console is clean" from "console UI is hiding everything."

Distinct from #861 (compile errors lost across domain reload) — this is filter inheritance, reproducible with plain `Debug.Log` in edit mode.

## Steps to Reproduce

1. In the Console window toolbar, toggle OFF the Log and Warning buttons (leave Error on)
2. `Debug.Log("PROBE")` (e.g. via `execute_code`)
3. Call `read_console` `action: "get"`, `types: ["log"]` (or omit types / use `"all"`)
4. **Result:** `{"success": true, "message": "Retrieved 0 log entries"}`
5. Toggle the buttons back on — the *same* call immediately returns the probe

## Root Cause

`ReadConsole.GetConsoleEntries` iterates `LogEntries.StartGettingEntries()` / `GetEntryInternal()`. Those internal APIs respect the Console window's `consoleFlags` (`LogLevelLog = 1<<7`, `LogLevelWarning = 1<<8`, `LogLevelError = 1<<9`) and its search filter (`SetFilteringText`). With the toggles off, `StartGettingEntries()` returns 0.

Verified by reflection in the failing session: `consoleFlags=7682` (LogLevelLog=false, LogLevelWarning=false, LogLevelError=true) -> 0 entries; after `LogEntries.SetConsoleFlag(1<<7, true)` and `(1<<8, true)` the identical `read_console` call returned all pending entries.

## Why it matters

The failure is invisible: the tool succeeds, so an agent concludes the console is clean — worst case right after a compile, where it will happily proceed on broken state. It cost us most of a debugging session before we found the toggle.

## Suggested fix

In `GetConsoleEntries`, snapshot `consoleFlags`, force the three `LogLevel*` bits on (and clear the filtering text) for the duration of the Start/End iteration, then restore. Alternatively (less invasive): when a requested type is masked by `consoleFlags`, include a warning in the response, e.g. `"note": "Console window filters are hiding log/warning entries"`.

## Environment

- MCP for Unity v10.0.0 (`com.coplaydev.unity-mcp@7b7db7b`)
- Unity 6000.3.9f1, Windows 11 (10.0.26200), HTTP transport

KamilDev:
## Description

`read_console` inherits the Console window's UI filters, so entries the user has hidden in the Editor are invisible to the agent — and the tool still reports `success: true`, so "Retrieved 0 log entries" is indistinguishable from a clean console. Worst case is right after a compile, where an agent proceeds on broken state.

`LogEntries` filtering state is global and shared with the Console window. `StartGettingEntries()` honors both:

- the toolbar's Log / Warning / Error severity toggles — `ConsoleFlags.LogLevelLog|Warning|Error`, bits `1<<7`, `1<<8`, `1<<9`
- the toolbar's search box — `LogEntries.SetFilteringText`

The tool's own `filterText` argument is applied client-side afterwards, so it cannot compensate for entries the native call never returned.

`GetConsoleEntries` now snapshots both, neutralizes them for the duration of the read, and restores them in the `finally` block once the iteration session is closed — so the user's Console view is left exactly as they had it. The tool's own `types` and `filterText` arguments still do the filtering, which is what the caller actually asked for.

## Type of Change

- [ ] Documentation update
- [x] Test update

## Changes Made

`MCPForUnity/Editor/Tools/ReadConsole.cs`

- Reflect `LogEntries.consoleFlags`, `SetFilteringText` and `GetFilteringText`.
- `TryForceLogLevelFlags` / `RestoreConsoleFlags` — OR the three severity bits on for the read, restore afterwards. Idempotent: when the bits are already set the property is not written at all.
- `TryClearFilteringText` / `RestoreFilteringText` — blank the search query for the read, restore afterwards.
- Restoration happens after `EndGettingEntries`, so it never runs inside an open iteration session, and it runs even if the iteration throws.

`TestProjects/UnityMCPTests/Assets/Tests/EditMode/Tools/ReadConsoleTests.cs`

- `HandleCommand_Get_IgnoresConsoleSearchFilter` — logs a probe, sets a non-matching search query, asserts the probe is still returned and that the query is left as the user set it.
- `HandleCommand_Get_IgnoresConsoleSeverityToggles` — logs a probe, switches the Log/Warning bits off, asserts the probe is still returned and that the flags are restored.

### Notes on the approach

- **The three severity bits only, not the whole mask.** `Collapse` (`1<<0`) also changes what `StartGettingEntries` returns, but forcing it off would flood a page with duplicates of a spammed message and push distinct entries past the `count`/`pageSize` limit — that hides output rather than revealing it. Left alone deliberately.
- **The search query is only cleared when `GetFilteringText` is also available**, so a user's search is never discarded without a way to put it back.
- **All four reflected members exist in 2021.3, 2022.3, 6000.0 and 6000.3** (checked against UnityCsReference at each tag), so this is not a version-gated path in practice. They are still reflected *optionally*: if a future Unity renames one, console reads degrade to the current behavior instead of failing outright. The `HandleCommand` reflection guard is deliberately unchanged — these members are not required for a console read to work at all, and `clear` does not depend on them.

- Unity version(s) tested: compile-verified against **2022.3.27f1** and **6000.3.14f1** for **win, osx and linux** each; behavior verified at runtime in a live **6000.3.14f1** Editor
- Package source used: local `file:` checkout
- Resolved commit hash from `Packages/packages-lock.json`: n/a (`file:` source)

- [ ] Unity EditMode tests
- [ ] Unity PlayMode tests

### Compile

`tools/compile-check.sh` run against local Hub installs of 2022.3.27f1 and 6000.3.14f1 — `MCPForUnity.Runtime` and `MCPForUnity.Editor` compile clean on win, osx and linux for both. The EditMode test assembly was compiled the same way (same reference set plus the built package DLLs) to confirm the two new tests build; the only diagnostics are pre-existing `CS0618`/`CS0649` warnings from unrelated test files.

### Runtime, in a live 6000.3.14f1 Editor

Verified against a project consuming this branch through a `file:` package reference, confirming first that the loaded `MCPForUnity.Editor` assembly actually contained the new methods.

**The bug, reproduced:** with the Log and Warning toggles switched off, a raw `LogEntries.StartGettingEntries()` returns **0** — with a freshly logged entry sitting in the console. That is the tool's old view.

**The fix, against that same state:** `read_console` returns the probe entry, and `consoleFlags` is left byte-identical afterwards.

**Search-box case** (the reported one): with the search box set to a query matching nothing, `read_console` returns the probe, and `GetFilteringText()` afterwards still returns the user's query unchanged.

**Idempotence confirmed incidentally:** in that project the three severity bits were already on, so `TryForceLogLevelFlags` correctly declined to write the property at all and `consoleFlags` never changed.

### On the two new tests

They have **not** been run as an NUnit fixture. `TestProjects/UnityMCPTests` is pinned to 2021.3.45f2 and I have no 2021 Editor installed; opening it in a newer one would silently upgrade the tracked `ProjectVersion.txt`. Note that the `Test in editmode` check on this PR passes in ~5s without a license, so it has not run them either — they need a licensed Editor run.

What I could do instead: execute both test bodies verbatim — same NUnit `Assert` calls, same setup and teardown — against the live Editor. Both pass. That exercises the assertion logic, but not the fixture plumbing (`[Test]` discovery, test-runner domain state), so a maintainer's licensed run is still the real gate.

Python tests were not run — this change touches no Python.

### Manual repro, for anyone verifying by hand

1. Type anything into the Console window's search box (or switch the Log/Warning toggles off).
2. `Debug.Log("PROBE")`.
3. `read_console` with `action: "get"`, `types: ["log"]`.
4. Before: `{"success": true, "message": "Retrieved 0 log entries"}`. After: the probe is returned, and the search box / toggles are left untouched.

## Documentation Updates

No tool or resource signature changed — the response shape and every argument are identical, so `website/docs/reference/` is unaffected.

## Related Issues

Fixes #1239

## Additional Notes

Supersedes #1252, which fixed the same issue and was closed unmerged by its author. Differences from that patch, all of them the review points it collected:

- The search query is saved and restored rather than cleared permanently — #1252 noted it cleared the box "without saving because `GetFilteringText` is not guaranteed to exist", but it does exist on every version this package supports, and losing a user's search on every console read is avoidable.
- Both mutations sit inside the same `try`/`finally` as the iteration, so a throw restores both.
- The `HandleCommand` reflection guard is left alone, so `clear` stays available regardless of these optional members.

Credit for the original diagnosis goes to @beast-ofcourse in #1239.

## Summary by CodeRabbit

* **Bug Fixes**
  * Console reads now include messages hidden by active search filters or disabled severity toggles.
  * The Console’s existing search and severity settings are preserved and restored after reading.
  * Console reads continue gracefully when these settings cannot be accessed, with warnings provided when necessary.
  * Restoration issues are reported without preventing the console read from completing.

### DIFF
--- a/MCPForUnity/Editor/Tools/ReadConsole.cs
+++ b/MCPForUnity/Editor/Tools/ReadConsole.cs
@@ -30,6 +30,19 @@ namespace MCPForUnity.Editor.Tools
         private static FieldInfo _messageField;
         private static FieldInfo _fileField;
         private static FieldInfo _lineField;
+
+        // Optional reflection members: used to neutralize the Console window's own filters
+        // while reading. Absent members degrade to the previous (filter-inheriting) behavior.
+        private static PropertyInfo _consoleFlagsProperty;
+        private static MethodInfo _setFilteringTextMethod;
+        private static MethodInfo _getFilteringTextMethod;
+
+        // Severity bits from the internal UnityEditor.ConsoleWindow.ConsoleFlags enum.
+        private const int ConsoleFlagLogLevelLog = 1 << 7;
+        private const int ConsoleFlagLogLevelWarning = 1 << 8;
+        private const int ConsoleFlagLogLevelError = 1 << 9;
+        private const int ConsoleFlagLogLevelMask =
+            ConsoleFlagLogLevelLog | ConsoleFlagLogLevelWarning | ConsoleFlagLogLevelError;
     
         // Static constructor for reflection setup
         static ReadConsole()
@@ -99,6 +112,13 @@ namespace MCPForUnity.Editor.Tools
                 if (_lineField == null)
                     throw new Exception("Failed to reflect LogEntry.line");
 
+                // Console window UI state. Present on every Unity version this package
+                // supports, but reflected optionally so a future rename degrades to the old
+                // behavior rather than disabling console reads outright.
+                _consoleFlagsProperty = logEntriesType.GetProperty("consoleFlags", staticFlags);
+                _setFilteringTextMethod = logEntriesType.GetMethod("SetFilteringText", staticFlags);
+                _getFilteringTextMethod = logEntriesType.GetMethod("GetFilteringText", staticFlags);
+
                 // (Calibration removed)
 
             }
@@ -115,6 +135,8 @@ namespace MCPForUnity.Editor.Tools
                     _getEntryMethod =
                         null;
                 _modeField = _messageField = _fileField = _lineField = null;
+                _consoleFlagsProperty = null;
+                _setFilteringTextMethod = _getFilteringTextMethod = null;
             }
         }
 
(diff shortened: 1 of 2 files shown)


Scriptwonder:
Hi there just got time to review this. Can you rebase this so I could take further look? Thanks!

KamilDev:
Rebased onto `beta` — now at c026cb49, mergeable.

The only conflict was in `ReadConsoleTests.cs`, against the tests dfb27f20 (#1348, the `LogEntry.mode` bit table) added at the same insertion point. Both sides are purely additive, so both are kept. `ReadConsole.cs` merged cleanly — that fix lives in `GetLogTypeFromMode` and the `ModeBit*` constants, which don't overlap the filter handling here.

Re-verified against a live Editor (6000.3.14f1) afterwards, since `read_console`'s `types` filter runs every entry through the newly corrected `GetLogTypeFromMode`:

- the mapping is the post-#1348 one on the rebased assembly (`0x804400 → Log`, `0x804200 → Warning`, `0x804100 → Error`);
- with a non-matching search query in the Console toolbar, the entry is still returned and the query is left as the user set it;
- with the Log and Warning severity toggles switched off, the entry is still returned and the toggles are left off.

As before, those two test bodies were executed against the Editor rather than run as an NUnit fixture — that still needs a licensed CI run.

Issue #1341, opened 2026-08-22
manage_build blocks on unsaved scene save dialog


TeapoyY:
## Problem

When `manage_build` triggers a player build while the open scene has unsaved changes, Unity shows a modal **Save modified scenes?** dialog. Because MCP runs headlessly/automated, the build hangs until a user dismisses the dialog manually.

## Expected behavior

`manage_build` should save assets and open scenes automatically before calling `BuildPipeline.BuildPlayer`, so automated builds are not blocked by save prompts.

## Steps to reproduce

1. Open a scene in the Unity Editor and make a change without saving.
2. Call `manage_build` with `action=build`.
3. Observe Unity prompts to save modified scenes and the build does not proceed until the dialog is handled.

## Proposed fix

Call `AssetDatabase.SaveAssets()` and `EditorSceneManager.SaveOpenScenes()` in `BuildRunner` immediately before `BuildPipeline.BuildPlayer` runs (same pattern already used by `TestRunnerService` for test runs).

## Environment

- Unity MCP `manage_build` tool
- Any Unity version where `BuildPipeline.BuildPlayer` prompts for unsaved scenes

TeapoyY:
Fix submitted in #1340.

PR #1340, opened 2026-08-21
fix(build): save assets and scenes before BuildPlayer

TeapoyY:
Fixes #1341

## Summary
- Call `AssetDatabase.SaveAssets()` and `EditorSceneManager.SaveOpenScenes()` in `BuildRunner` before `BuildPipeline.BuildPlayer` runs
- Prevents `manage_build` from blocking on Unity's unsaved-scene save dialog when scenes or assets were modified

## Test plan
- [ ] Open a scene, modify it without saving, then run `manage_build` with action `build`
- [ ] Confirm the build starts without a save popup
- [ ] Confirm assets with unsaved changes are persisted before the build completes

PR #527, opened 2026-01-07
Fix issue #525: Save dirty scenes for all test modes

sjennings:
## Summary
- Moved `SaveDirtyScenesIfNeeded()` call outside the PlayMode conditional so EditMode tests don't get blocked by Unity's "Save Scene" modal dialog

## Problem
When running EditMode tests through MCP with unsaved scene changes, Unity displays a "Save Scene" modal dialog that blocks the editor, causing MCP to timeout without receiving a response.

## Solution
The `SaveDirtyScenesIfNeeded()` method was only being called for PlayMode tests. This one-line change ensures dirty scenes are saved before ALL test runs, not just PlayMode.

## Testing
- Verified EditMode tests no longer block on save dialogs
- 1165 tests run successfully with this fix

Fixes #525

## Summary by Sourcery

Bug Fixes:
- Prevent EditMode test runs from being blocked by Unity's save scene modal dialog by saving dirty scenes for all test modes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where EditMode tests were blocked by save dialogs. Dirty scenes are now automatically saved before running any test mode, ensuring smooth test execution across all test types.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

Issue #525, opened 2026-01-07
EditMode tests blocked by "Save Scene" dialog when scene is dirty


sjennings:
## Summary

  When running EditMode tests via MCP with a dirty scene, Unity's "Save Scene" modal dialog blocks the editor event loop, causing MCP to timeout without receiving a response.

  ## Root Cause

  In `TestRunnerService.RunTestsAsync()`, the `SaveDirtyScenesIfNeeded()` method is only called for PlayMode tests:

  ```csharp
  // Current code (line 109-112)
  if (mode == TestMode.PlayMode)
  {
      SaveDirtyScenesIfNeeded();
  }
```

  This means EditMode tests don't auto-save dirty scenes, and Unity's test framework triggers a modal save dialog that blocks the entire editor.

## Steps to Reproduce

  1. Open a Unity project with MCP-for-Unity installed
  2. Make changes to the current scene (scene becomes dirty)
  3. Call run_tests with mode: "EditMode" via MCP
  4. Observe: Unity shows "Save Scene" modal dialog
  5. Result: MCP times out waiting for response

## Expected Behavior

  EditMode tests should auto-save dirty scenes before running, just like PlayMode tests do.

## Proposed Fix

  Move SaveDirtyScenesIfNeeded() outside the PlayMode conditional so it runs for all test modes:

  ```csharp
  // Before _testRunnerApi.Execute(settings):
  SaveDirtyScenesIfNeeded();  // Save for ALL test modes, not just PlayMode
```

  This is a one-line change in TestRunnerService.cs.

##  Environment

  - Unity MCP version: 8.7.1
  - Unity version: 6.3 LTS
  - MCP Client: Claude Code

##  Workaround

  Manually save all scenes before running tests, or create a [InitializeOnLoad] script that registers a ICallbacks handler to auto-save on RunStarted.

PR #1350, opened 2026-08-28
feat: report and answer modal dialogs that block the Editor

KamilDev:
## Description

A modal dialog stalls Unity's main thread, so `EditorApplication.update` stops firing and every queued MCP command times out. Today the bridge reports that exactly like an ordinary busy Editor:

```
Unity did not respond to 'read_console' within 2.0s; please retry
Unity session not ready for 'get_editor_state' (ping not answered); please retry
Refresh triggered but timed out after 60s waiting for editor readiness.
```

Every one says `hint: "retry"`. So the agent's correct-looking behaviour — poll a few times, then hand the task back — is exactly wrong: it burns turns and reports a false conclusion, when one sentence naming the dialog would have resolved it. The Editor stays wedged until a human walks over and clicks a button.

This makes a main-thread stall a distinguishable state, names the dialog, and lets the caller answer it.

## Type of Change

- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change
- [ ] Documentation update
- [ ] Test update

## Changes Made

**Detection.** `MainThreadHeartbeat` stamps a timestamp from `EditorApplication.update`. A new `liveness` command is answered on the transport receive thread instead of the main-thread queue, returning the stall, the dispatcher queue depth, and any modal.

The heartbeat must be written by the main thread. A heartbeat stamped by whichever thread answers the request only proves that thread ran, and reports a healthy Editor while the main thread is frozen.

**Identification (Windows).** The probe keys on the modal's *owner window being disabled* — what a modal is at the OS level — rather than on a window class, so it covers both kinds:

| | `EditorUtility.DisplayDialog` | `EditorWindow.ShowModal` |
|---|---|---|
| Window | native `#32770` | Unity-drawn `UnityContainerWndClass` |
| Title / body / buttons | readable | title only |
| Answerable | yes, real `Button` controls | no |

The window scan runs on a sampler thread and never on the request path: `GetWindowText` on a window owned by the calling process is a synchronous `WM_GETTEXT`, which blocks when the main thread is not pumping — the exact case being reported on. Measured: an unguarded read hung the probe for 10s.

**Reporting.** Stalls are classified as `modal_dialog` or `main_thread_blocked`, with `hint` of `answer_dialog`, `user_action_required`, or `wait` instead of a bare `retry`. `wait_for_editor_ready` stops polling on a modal rather than spending its full 60s on a state that cannot clear on its own. A plugin that does not answer `liveness` falls back to today's behaviour, so no version handshake is needed.

**Answering.** `answer_dialog` reads the dialog, or presses a button by label. Routed off the main thread so it does not queue behind the dialog it is clearing, and refuses if the open dialog no longer matches the title the caller was shown.

**Prevention.** `refresh_unity` reconciles an open scene changed on disk *before* `AssetDatabase.Refresh()`, so the reload prompt has no reason to appear. It reloads only when nothing is lost and otherwise refuses — both answers discard something, so the caller chooses via `on_external_scene_change`.

**Also: the websocket receive loop now handles messages detached.** This is the one behaviour change beyond the feature and the reason the rest works. Awaiting `HandleMessageAsync` inline meant one command waiting on the main thread stopped every later frame from being *read at all* — including the liveness probe and the server's pings. Execution is still ordered by `TransportCommandDispatcher`'s queue and sends are still serialised by `_sendLock`, so only the read side becomes concurrent.

- Unity version(s) tested: **6000.3.14f1**
- Package source used: `file:` (local checkout)
- Resolved commit hash from `Packages/packages-lock.json`: n/a (local `file:` source)

Detection is Windows-only. macOS/Linux report the stall without naming the dialog and fall back to `user_action_required`; the heartbeat and classification are cross-platform. No new `#if UNITY_*` gates; platform is checked at runtime.

- [x] Python tests (`cd Server && uv run pytest tests/ -v`) — **1385 passed, 3 skipped**
- [x] Unity EditMode tests — **10/10 passed** via the Test Runner
- [ ] Unity PlayMode tests — not applicable, no runtime surface

Reproduced against the genuine Unity prompt, not a synthetic one. With the Editor wedged, an ordinary command still fails at 2.0s while the new path answers in **0.32s**:

```
title:   The open scene(s) have been modified externally
body:    The following open scene(s) have been changed on disk:
          Assets/Scenes/<scene>.unity
         Do you want to reload the scene(s)?
buttons: Reload, Ignore
main_thread_stall_ms: 7327   pending_commands: 1
```

Then `answer_dialog(button="Reload")` unblocked the Editor in 0.33s. Also verified: the title-mismatch guard refuses, an unknown button lists the real options, `ShowModal` is reported as blocked but not answerable, and a long non-modal main-thread operation stays `hint: "wait"` rather than crying wolf.

Stdio is compile-verified only; HTTP was the live path.

## Documentation Updates

- [x] I have added/removed/modified tools or resources
- [x] If yes, I have updated all documentation files using:
  - [ ] The LLM prompt at `tools/UPDATE_DOCS_PROMPT.md`
  - [x] Manual review of the generated changes

`tools/generate_docs_reference.py` regenerated; only the genuinely changed pages are included.

## Related Issues

Relates to #1341 — same class of problem, approached generally rather than per-dialog. #525 / #527 and #1340 each fix one dialog by saving before the operation; that cannot cover a prompt raised by project or third-party editor code, or one Unity raises on its own schedule.

## Additional Notes

Worth a reviewer's attention:

1. **The detached receive loop** is the change most likely to have consequences beyond this feature.
2. **`answer_dialog` presses buttons.** It is deliberately not automatic — nothing answers a dialog unless a caller names a button. `WM_CLOSE` would also close a Unity-drawn modal, but that is a blind cancel with no readable options, so it is not exposed.
3. Modal onset is **not** correlated with the call that caused it — during testing a dialog surfaced ~3 minutes after the `delayCall` that scheduled it, with no command in flight. That is why detection is continuous rather than per-call.

## Summary by CodeRabbit

* **New Features**
  * Added support for detecting and answering blocking Unity Editor dialogs.
  * Added the `answer_dialog` tool and `editor dialog` command.
  * Added editor liveness reporting to distinguish modal blocks from main-thread stalls.
  * Added configurable handling for scenes changed externally during refresh.
* **Bug Fixes**
  * Commands and status checks remain responsive when Unity’s main thread is blocked.
  * Refresh now reports blocked dialogs and external scene changes instead of generic timeouts.
* **Documentation**
  * Added references for dialog handling and external scene-change options.

### DIFF
--- a/MCPForUnity/Editor/Helpers/ModalDialogProbe.cs
+++ b/MCPForUnity/Editor/Helpers/ModalDialogProbe.cs
@@ -1,0 +1,297 @@
+using System;
+using System.Collections.Generic;
+using System.Runtime.InteropServices;
+using System.Text;
+using UnityEngine;
+
+namespace MCPForUnity.Editor.Helpers
+{
+    /// <summary>
+    /// Describes a native modal dialog blocking the Editor's main thread.
+    /// </summary>
+    internal sealed class ModalDialogInfo
+    {
+        public bool Supported { get; set; }
+        public bool Blocked { get; set; }
+
+        /// <summary>
+        /// "dialog" for a native <c>EditorUtility.DisplayDialog</c> box, whose buttons are real OS
+        /// controls; "editor_window" for a Unity-drawn modal (<c>EditorWindow.ShowModal</c>), which
+        /// blocks identically but paints its own buttons with IMGUI.
+        /// </summary>
+        public string Kind { get; set; }
+
+        /// <summary>Whether a specific button can actually be pressed programmatically.</summary>
+        public bool Answerable { get; set; }
+
+        public long Handle { get; set; }
+        public string Title { get; set; }
+        public string Body { get; set; }
+        public List<string> Buttons { get; } = new List<string>();
+    }
+
+    /// <summary>
+    /// Reads (and can answer) the native modal dialog that blocks Unity's main thread.
+    ///
+    /// Windows only. <c>EditorUtility.DisplayDialog</c> raises a standard Win32 dialog (class
+    /// <c>#32770</c>) owned by the Editor process, whose title, body and button labels are all
+    /// readable, and whose buttons are real <c>Button</c> controls that accept <c>BM_CLICK</c>.
+    ///
+    /// Every text read goes through <see cref="SendMessageTimeoutW"/>: <c>GetWindowText</c> on a
+    /// window owned by the calling process is a synchronous <c>WM_GETTEXT</c>, which hangs when the
+    /// owning thread is not pumping messages — exactly the case this probe reports on.
+    /// </summary>
+    internal static class ModalDialogProbe
+    {
+        private const string DialogWindowClass = "#32770";
+        private const string ButtonWindowClass = "Button";
+        private const uint WM_GETTEXT = 0x000D;
+        private const uint BM_CLICK = 0x00F5;
+        private const uint SMTO_ABORTIFHUNG = 0x0002;
+        private const int TextReadTimeoutMs = 400;
+        private const int MaxTopLevelScan = 5000;
+        private const int MaxChildScan = 200;
+
+        [DllImport("user32.dll", CharSet = CharSet.Unicode)]
+        private static extern IntPtr FindWindowExW(IntPtr parent, IntPtr childAfter, IntPtr className, IntPtr windowName);
+
+        [DllImport("user32.dll")]
+        private static extern uint GetWindowThreadProcessId(IntPtr hWnd, out uint processId);
+
+        [DllImport("user32.dll")]
+        private static extern bool IsWindowVisible(IntPtr hWnd);
+
+        [DllImport("user32.dll")]
+        private static extern bool IsWindowEnabled(IntPtr hWnd);
+
+        [DllImport("user32.dll")]
+        private static extern bool IsWindow(IntPtr hWnd);
+
+        [DllImport("user32.dll", CharSet = CharSet.Unicode)]
+        private static extern int GetClassNameW(IntPtr hWnd, StringBuilder buffer, int maxCount);
+
+        [DllImport("user32.dll", CharSet = CharSet.Unicode)]
+        private static extern IntPtr SendMessageTimeoutW(
+            IntPtr hWnd, uint msg, IntPtr wParam, StringBuilder lParam, uint flags, uint timeoutMs, out IntPtr result);
+
+        [DllImport("user32.dll")]
+        private static extern bool PostMessageW(IntPtr hWnd, uint msg, IntPtr wParam, IntPtr lParam);
+
+        internal static bool IsSupported =>
+            Application.platform == RuntimePlatform.WindowsEditor;
+
+        /// <summary>
+        /// Look for a modal dialog owned by this process. Safe to call from a background thread.
+        /// </summary>
+        internal static ModalDialogInfo Capture()
+        {
+            var info = new ModalDialogInfo { Supported = IsSupported };
+            if (!info.Supported)
+            {
+                return info;
+            }
+
+            try
+            {
+                uint myPid = (uint)System.Diagnostics.Process.GetCurrentProcess().Id;
+
+                // Windows disables a modal's owner while the modal is up. That — not the window
+                // class — is what "a modal is blocking the Editor" means, and it is what makes a
+                // Unity-drawn EditorWindow.ShowModal visible here too.
+                var enabled = new List<IntPtr>();
+                bool anyDisabled = false;
+
+                IntPtr window = IntPtr.Zero;
+                for (int i = 0; i < MaxTopLevelScan; i++)
+                {
+                    window = FindWindowExW(IntPtr.Zero, window, IntPtr.Zero, IntPtr.Zero);
+                    if (window == IntPtr.Zero)
+                    {
+                        break;
+                    }
+
+                    GetWindowThreadProcessId(window, out uint pid);
+                    if (pid != myPid || !IsWindowVisible(window))
+                    {
+                        continue;
+                    }
+
+                    if (IsWindowEnabled(window))
+                    {
+                        enabled.Add(window);
+                    }
+                    else
+                    {
+                        anyDisabled = true;
+                    }
+                }
+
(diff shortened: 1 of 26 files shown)


KamilDev:
Thanks — five of the seven landed in be6d84a3. Two I'm pushing back on, with evidence.

**Fixed**

- **`answerable` reads the wrong field** — correct, and the most valuable catch here. On Windows an `EditorWindow.ShowModal` reports `supported=true, answerable=false`, so the read path promised a press that Unity always refuses. That is precisely the confusion the separate field exists to prevent. Fixed, plus a regression test verified to fail on the old behaviour.
- **Baseline not initialised on scene open** — correct and the most consequential. With no baseline the first refresh of a session recorded the already-changed mtime and let the edit through to `AssetDatabase.Refresh`, so the guard did nothing on first use. Now recorded on `sceneOpened`, `sceneSaved` and at load. Verified live: with only a load-time baseline and no prior refresh, an external edit is now detected and the scene reloaded.
- **`answer_dialog` timeout budget** — agreed, and worse than a slow path: the click is posted rather than awaited, so expiring mid-answer would report failure for a dialog that was in fact answered. Given its own 10s budget.
- **Handler runs inline on the receive loop** — right in effect, though not for `liveness`: that reads a cached snapshot and never scans windows. `answer_dialog` does scan, so it could hold the loop — the very thing detaching the loop was meant to prevent. Now on the thread pool.
- **Self-validating test** — correct, it asserted the literals it had just assigned and could not fail. Removed; the invariant is covered by the `describe_stall` tests.

**Not applying**

- **`answer_dialog.py:47` — read `result.data` before accessing `modal`.** Both transports already unwrap. HTTP passes through `normalize_unity_response`, which returns `result` when it contains `success` (`models/unity_response.py:21-22`); the stdio path returns `resp.get('result', {})` (`legacy/unity_connection.py:430`). Every other tool in the repo reads `response.get("data")` on this same path.
- **`StdioBridgeHost.cs:608` — authorization bypass (CWE-862).** There is no gate being bypassed. The dispatcher only enforces enablement when metadata exists: `if (toolMeta != null && !IsToolEnabled(...))`. Checked at runtime — `GetToolMetadata("answer_dialog")` returns `null`, so routing through the dispatcher would not gate it either. These are transport-level commands like `ping`, which is likewise answered inline in both transports without a tool gate; enablement is enforced server-side, before the command reaches the plugin. Happy to add a check if maintainers would rather they be registered tools, but today it would be a no-op.

Re-verified after the changes: Python 1386 passed / 3 skipped, EditMode 9/9 via the Test Runner, and the end-to-end flow re-run against the real "The open scene(s) have been modified externally" prompt.

KamilDev:
Round 2 — all three applied in bdcc7572. Two of them were bugs I introduced while fixing round 1, which is a fair hit.

- **Baselines for unrelated open scenes** — correct, and the more serious of the two. Recording on scene events was round 1's fix, but doing it as a full map rebuild meant opening or saving scene B stamped a fresh mtime onto scene A, so A's pending edit became its own baseline and `Reconcile` stopped reporting it. The refresh then walked into the modal the guard exists to prevent. Callbacks now update only the scene they name, and the delayed initialisation fills missing entries without overwriting existing ones — an existing entry may be the only record of an unreconciled edit.

- **Liveness transport errors swallowed** — correct. A failed probe carries no modal, so an unavailable session came back as a confident "No modal dialog is currently open". Same failure shape as the `answerable` bug: a wrong answer stated with certainty. Now propagates the unsuccessful response, with a test verified to fail on the old behaviour.

- **Unbounded detached `execute` admission** — agreed, and it is a regression I introduced: awaiting the receive loop used to provide backpressure, and detaching it removed that with nothing in its place. A burst can now queue unboundedly while the main thread works through it one command at a time. Added a bounded gate (16) around dispatcher-bound commands only. Deferring rather than rejecting keeps command semantics unchanged and the server's own timeout still bounds the caller.

  `liveness` and `answer_dialog` are deliberately outside that gate — a saturated gate is precisely one of the states they exist to report on, and putting them behind it would recreate the original bug in a new place. Verified live: with a dispatcher-bound command blocked and holding a slot, `liveness` still answers in 0.33s while `read_console` fails at 2.0s.

Re-verified: Python 1387 passed / 3 skipped, EditMode 9/9 via the Test Runner, end-to-end re-run against a live blocking dialog.

On the codecov patch-coverage note: the uncovered lines are concentrated in `cli/commands/editor.py`, which has no test coverage anywhere in this repo, and in the `plugin_hub` branches that only execute against a genuinely blocked Editor. The classification logic itself is unit-tested, and the blocked paths were exercised live rather than mocked. Happy to add CLI tests if maintainers want that gap closed here rather than separately.

KamilDev:
Round 3 — the timeout finding is correct and was a real defect; fixed in cad20413. The second I've answered rather than applied, because I think the first fix subsumes it.

**Timeout must cover admission** — right, and worse than a queueing nicety. The gate waited on the *connection* token, so a command could sit indefinitely and only start its timeout once admitted. Past a backlog that means executing for a caller that already gave up: side effects applied for nobody. The timeout is now armed before the wait and its token passed to both the wait and the dispatch, so a command that cannot get in within its own budget expires with the existing timeout response instead of running late.

**Hard outstanding-command limit with rejection** — I've not added a separate cap, because the fix above already bounds this and does it on a better axis. Waiting handlers now hold their payloads only until their own deadline, so outstanding work is bounded by arrival rate × the caller's timeout rather than growing without limit; that was the substance of the concern. A fixed count cap on top would reject commands that are within budget and would have succeeded, trading a real failure for a hypothetical one, and it needs a magic number that cannot be derived from anything the plugin knows. If maintainers would rather have a hard ceiling regardless, it's a small change and I'll add it — I just don't want to pick an arbitrary limit unprompted.

Worth noting for the record that both round-2 and round-3 findings were on code I added in the previous round's fixes, not in the original PR. The gate exists because detaching the receive loop removed backpressure; arming the timeout correctly is what makes that gate sound.

Re-verified after the change: Python 1387 passed / 3 skipped, EditMode 9/9 via the Test Runner, and the live path re-checked — with a command blocked and holding a slot, `liveness` still answers in 0.35s while `read_console` fails at 2.0s, and `answer_dialog` clears it.

Issue #1134, opened 2026-05-18
AssetImportWorker crash in CommandRegistry.AutoDiscoverCommands during domain reload


jonlepage:
The AssetImportWorker crashes during domain reload when CommandRegistry
scans assemblies via reflection in its [InitializeOnLoad] static constructor.

The crash occurs at CommandRegistry.cs:72 inside AutoDiscoverCommands(),
when calling GetCustomAttribute<T>() on a type that is not fully loaded
in the worker process context. This triggers a native crash in
mono_class_from_mono_type_internal (access violation).

This happens intermittently during script recompilation and can destabilize
the main Editor process (freezes, force-kill required).

Actual result:
- Editor freezes (stops responding to Windows) or crashes with access violation
- Windows Event Viewer shows:
  Application Error: Unity.exe 6000.4.1.13702
  Faulting module: unknown (0.0.0.0)
  Exception code: 0xc0000005 (STATUS_ACCESS_VIOLATION)
  Fault offset: 0x0000028e84cdf218
- No crash stack trace in Editor.log (log ends with normal scene unload)
- "Saving Failed" dialog sometimes appears: "You are trying to save a Prefab with a missing script"
- After force-kill: "Prefab Has Been Changed on Disk" dialog on relaunch
- 3 incidents in single session confirmed via Windows Application Event Log:
  17:53 - Application Error (crash 0xc0000005)
  15:25 - Application Hang (freeze, force-killed)
  14:48 - Application Hang (freeze, force-killed)

Stack trace origin: Logs/AssetImportWorker2.log
Unity version: 6000.4.1f1
Plugin version: com.coplaydev.unity-mcp@13fb3ee12774
```bat
The AssetImportWorker2 has crashed. Please report a bug with the following Stack Trace:
========== OUTPUTTING STACK TRACE ==================

0x00007FF94394AA4E (mono-2.0-bdwgc) [C:\build\output\Unity-Technologies\mono\mono\metadata\class.c:2169] mono_class_from_mono_type_internal 
0x00007FF9439C7C9A (mono-2.0-bdwgc) [C:\build\output\Unity-Technologies\mono\mono\metadata\icall-def.h:921] ves_icall_RuntimeTypeHandle_GetAttributes_raw 
0x000001CC4084E096 (Mono JIT Code) (wrapper managed-to-native) System.RuntimeTypeHandle:GetAttributes (System.RuntimeType)
[...]

PR #1358, opened 2026-09-01
perf: discover tools and resources via TypeCache instead of scanning every assembly

### LINKED ISSUE #1336: [Bug]: CommandRegistry.AutoDiscoverCommands adds ~9s to every domain reload in large projects
### What happened?

On a large commercial Unity project (several hundred assemblies), every domain reload spends **~8.9 s inside `CommandRegistry.AutoDiscoverCommands()`** — roughly half of our total ~20 s reload time. Since the registry re-initializes after every reload, we pay this cost on every script compile and every Play Mode entry.

The cause is that discovery enumerates **every type of every loaded assembly** (`UnityAssembliesCompat.GetLoadedAssemblies()` → `GetTypes()`) and then checks `GetCustomAttribute<T>()` per type, twice (tools + resources). The cost scales with the project's total type count, not with the number of MCP tools.

`ToolDiscoveryService.DiscoverAllTools()` has the same issue: it runs a TypeCache primary scan but then **unconditionally** runs the exhaustive AppDomain scan as a fallback and merges the results, so the full-domain scan cost is paid there as well.

**Expected:** discovery should use `UnityEditor.TypeCache.GetTypesWithAttribute<T>()`, which is a prebuilt-index lookup and effectively free. The codebase already trusts TypeCache for exactly this pattern — `ResourceDiscoveryService` is TypeCache-only, and `McpClientRegistry`, `GetMenuItems`, and the graphics helpers all use TypeCache. `CommandRegistry` (and the ToolDiscoveryService fallback) are the remaining exhaustive scans.

Side benefit: TypeCache reads a prebuilt native index instead of reflecting over assembly metadata, which avoids the half-loaded-metadata crash class that led to the AssetImportWorker workaround in #1134.

Regarding the existing comment in `ToolDiscoveryService` ("TypeCache … can miss project assemblies in some domain-reload states"): if there is a known repro for that state we would love details. Otherwise a conservative fix is TypeCache primary + AppDomain fallback **only when TypeCache returns nothing** — same coverage guarantees, none of the per-reload cost.

**Happy to submit a PR** (against `beta`) for both call sites either way — just let us know which shape you prefer (TypeCache-only vs. lazy fallback).

### Reproduction steps

1. Install `com.coplaydev.unity-mcp` 10.1.0 in a project with a large number of assemblies/types (we see it with several hundred asmdefs; smaller projects pay proportionally less).
2. Trigger a domain reload (edit any script and recompile, or enter Play Mode).
3. Profile the reload — `CommandRegistry.AutoDiscoverCommands` accounts for ~8.9 s of it.
4. Replacing the scan with `TypeCache.GetTypesWithAttribute<McpForUnityToolAttribute>()` / `TypeCache.GetTypesWithAttribute<McpForUnityResourceAttribute>()` finds the same tools/resources with discovery time effectively at 0.

### Unity version

2021.3.56f2

### MCP for Unity package version

10.1.0 (OpenUPM)

### Python server version

n/a — editor-side issue; the Python server is not involved.

### MCP client

Claude Code

### Transport

stdio (the scan runs on every domain reload regardless of transport)

### OS

Windows 11

### Relevant logs / console output

Editor profiling of a domain reload shows ~8.9 s attributed to `CommandRegistry.AutoDiscoverCommands()` out of a ~20 s total reload. Happy to attach profiler captures if useful.

### Checks

- [x] I searched existing issues and did not find a duplicate
- [x] I included logs / steps to reproduce

ananttheant:
## Description

`CommandRegistry.AutoDiscoverCommands` walks every type of every loaded assembly and then runs two full attribute-matching passes over that list, once per domain reload. `ToolDiscoveryService.DiscoverAllTools` is worse: it queries `TypeCache` first and then runs the same exhaustive `AppDomain` walk anyway — unconditionally, merging and de-duplicating the two results — so the fast path never actually saved anything.

`ResourceDiscoveryService` already trusts `TypeCache` alone, so this brings the other two in line with it.

Measured in this repo's own `TestProjects/UnityMCPTests` on Unity 2022.3.62f1 — 193 non-dynamic assemblies, 44,195 types, averaged over 5 runs after warm-up:

| Path | Time | Types found |
|---|---|---|
| `TypeCache` | 0.01 ms | 54 |
| Exhaustive assembly walk | 310.63 ms | 54 |

Both find **exactly the same 54 types**, which is the important part: this is a pure cost reduction, not a change in what gets discovered. That is 310 ms per domain reload in a near-empty project; the multi-second figure in #1336 comes from real projects carrying far more assemblies.

## Type of Change
- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Documentation update
- [x] Refactoring (no functional changes)
- [x] Test update

## Changes Made

- `CommandRegistry.AutoDiscoverCommands` reads tools and resources from `TypeCache`. The old assembly walk moves into a `GetAllLoadedTypes()` helper, reached only when the index returns nothing at all for this domain.
- `ToolDiscoveryService.DiscoverAllTools` runs the `AppDomain` scan only when `TypeCache` came back empty, instead of always running it and merging.
- New `CommandRegistry.ShouldFallBackToExhaustiveScan(int, int)` pure predicate expressing that rule, following the existing `ShouldAbandonBusyPort` / `ShouldKeepWaitingForReady` pattern, plus an EditMode test covering all four combinations.
- The AssetImportWorker guard from #1134 is untouched and still short-circuits before any scanning.

**The one behavioural trade-off worth your call:** the old code always merged the exhaustive scan in, so a type that `TypeCache` somehow missed while other types resolved would still have been found. Now a non-empty index is treated as authoritative. I chose this because it keeps a safety net for the "index not ready" case the original comments describe while removing the cost from the normal path — but if you would rather always merge, or gate the fallback differently (say, per-attribute rather than on the combined count), that is a small change to `ShouldFallBackToExhaustiveScan` and its call site.

- Unity version(s) tested: 2022.3.62f1 (full EditMode suite + benchmark)
- Package source used: `file:` (local clone, branch `perf/typecache-command-discovery` off `beta`)
- Resolved commit hash from `Packages/packages-lock.json`: n/a (file reference)

`TypeCache` is `UnityEditor` API available well below this package's 2021.3 floor, so no version gating is needed.

- [ ] Python tests — not applicable, no Python-side changes
- [x] Unity EditMode tests
- [ ] Unity PlayMode tests — not applicable, editor-only services

Full EditMode suite on 2022.3.62f1: **1121 passed, 0 failed** (66 pre-existing ignores). The existing `CommandRegistryTests.AutoDiscovery_RegistersAllBuiltInTools` and `ToolDiscoveryServiceTests` are the real regression guard here — they assert the built-in tools still resolve, and they pass unchanged.

## Documentation Updates
- [ ] I have added/removed/modified tools or resources — no tool or resource surface changed

## Related Issues

Closes #1336

## Additional Notes

The benchmark above came from a throwaway test I did not commit; it timed both strategies in the same editor session and asserted their results matched. Happy to add it as a permanent `[Explicit]` test if that is useful to you, though wall-clock assertions in CI tend to be flaky, which is why I left it out.

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved tool, command, and resource discovery speed in the Unity Editor by prioritizing indexed discovery results.
  * Reduced unnecessary scanning of loaded assemblies during setup and refresh operations.
* **Reliability**
  * Added a fallback scan when indexed discovery finds no tools or resources.
  * Improved handling of assemblies that cannot be inspected, helping discovery complete more reliably.

### DIFF
--- a/MCPForUnity/Editor/Services/ToolDiscoveryService.cs
+++ b/MCPForUnity/Editor/Services/ToolDiscoveryService.cs
@@ -23,28 +23,35 @@ namespace MCPForUnity.Editor.Services
 
             _cachedTools = new Dictionary<string, ToolMetadata>();
 
-            // Primary scan via TypeCache (fast, but can miss project assemblies in some domain-reload states)
-            var toolTypes = TypeCache.GetTypesWithAttribute<McpForUnityToolAttribute>();
+            // Primary scan via TypeCache: answers from Unity's prebuilt attribute index instead of
+            // walking every type of every loaded assembly, which the exhaustive scan below did on
+            // every call regardless of what the index already found (issue #1336).
+            TypeCache.TypeCollection toolTypes = TypeCache.GetTypesWithAttribute<McpForUnityToolAttribute>();
 
-            // Fallback scan via AppDomain (slower but exhaustive; mirrors CommandRegistry behaviour)
-            var appDomainTypes = AppDomain.CurrentDomain.GetAssemblies()
-                .Where(a => !a.IsDynamic)
-                .SelectMany(a =>
-                {
-                    try { return a.GetTypes(); }
-                    catch (Exception ex)
+            List<Type> allToolTypes;
+            if (toolTypes.Count > 0)
+            {
+                allToolTypes = toolTypes.ToList();
+            }
+            else
+            {
+                // Fallback scan via AppDomain (slower but exhaustive), reached only when the index
+                // is not usable for this domain yet; mirrors CommandRegistry behaviour.
+                allToolTypes = AppDomain.CurrentDomain.GetAssemblies()
+                    .Where(a => !a.IsDynamic)
+                    .SelectMany(a =>
                     {
-                        McpLog.Warn($"Failed to reflect types from assembly {a.FullName}: {ex.Message}");
-                        return new Type[0];
-                    }
-                })
-                .Where(t => t.GetCustomAttribute<McpForUnityToolAttribute>() != null);
-
-            // Merge both scans, deduplicating by type
-            var allToolTypes = toolTypes
-                .Concat(appDomainTypes)
-                .Distinct()
-                .ToList();
+                        try { return a.GetTypes(); }
+                        catch (Exception ex)
+                        {
+                            McpLog.Warn($"Failed to reflect types from assembly {a.FullName}: {ex.Message}");
+                            return new Type[0];
+                        }
+                    })
+                    .Where(t => t.GetCustomAttribute<McpForUnityToolAttribute>() != null)
+                    .Distinct()
+                    .ToList();
+            }
 
             foreach (var type in allToolTypes)
             {
--- a/MCPForUnity/Editor/Tools/CommandRegistry.cs
+++ b/MCPForUnity/Editor/Tools/CommandRegistry.cs
@@ -89,9 +89,15 @@ namespace MCPForUnity.Editor.Tools
                     resourceTypes = allTypes.Where(t => HasAttributeSafe<McpForUnityResourceAttribute>(t));
                 }
 
+                // Ordered by FullName because RegisterCommandType lets a duplicate command
+                // name overwrite the previously registered handler, so registration order
+                // decides which type wins a collision. Neither TypeCache nor the fallback
+                // documents an order, so without sorting a name collision could resolve
+                // differently between domain reloads.
+
                 // Discover tools
                 int toolCount = 0;
-                foreach (var type in toolTypes)
+                foreach (var type in toolTypes.OrderBy(t => t.FullName, StringComparer.Ordinal))
                 {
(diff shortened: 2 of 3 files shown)


ananttheant:
Superseded by #1364, which landed the same TypeCache switch — closing this.

Yours is the better change: you dropped the exhaustive fallback entirely rather than keeping it for the empty-index case, and the `OrderBy(t => t.FullName)` is a genuine catch I missed. `RegisterCommandType` lets a duplicate command name overwrite the previous handler, so registration order decides the winner, and TypeCache documents no order — my version would have made collisions resolve differently between reloads. Good spot.

For whatever it is worth, the numbers I measured in `TestProjects/UnityMCPTests` (193 assemblies, 44,195 types, Unity 2022.3.62f1) back the change up: 310.63 ms for the assembly walk versus 0.01 ms for TypeCache, both finding exactly the same 54 types.

PR #1043, opened 2026-04-06
feat: add component reference wiring — get_referenceable, set_reference, batch_wire

zaferdace:
## Summary

Extends `manage_components` with 3 new actions for automating object reference field assignment — the programmatic equivalent of drag-and-drop in the Inspector.

### New Actions

| Action | What it does |
|--------|-------------|
| `get_referenceable` | List scene objects and project assets assignable to a reference field |
| `set_reference` | Assign an object reference to a component field |
| `batch_wire` | Wire multiple references in one atomic call |

### Example Usage

```python
# "What can I assign to the Player's AudioSource.clip field?"
manage_components(
    action="get_referenceable",
    target="Player",
    component_type="AudioSource",
    property="clip"
)

# "Set the Main Camera's target to the Player transform"
manage_components(
    action="set_reference",
    target="MainCamera",
    component_type="CinemachineCamera",
    property="Follow",
    reference_path="Player"
)

# "Wire up all references on the GameManager at once"
manage_components(
    action="batch_wire",
    target="GameManager",
    component_type="GameManager",
    references=[
        {"property_name": "player", "reference_path": "Player"},
        {"property_name": "scoreUI", "reference_path": "Canvas/ScoreText"},
        {"property_name": "bgMusic", "reference_asset_path": "Assets/Audio/BGM.mp3"},
    ]
)
```

### Implementation Details

**Python** (`Server/src/services/tools/manage_components.py`):
- Added new parameters: `reference_path`, `reference_asset_path`, `reference_instance_id`, `references`, `clear`, `atomic`, `include_scene`, `include_assets`, `limit`
- Routes through `send_with_unity_instance` same as existing actions

**C#** (`MCPForUnity/Editor/Tools/ManageComponents.cs`):
- `get_referenceable`: resolves field type via reflection, searches scene (`FindObjectsOfType`) and assets (`AssetDatabase.FindAssets`) for matching types
- `set_reference`: uses `SerializedObject` + `SerializedProperty.objectReferenceValue` for correct prefab override tracking and Undo support
- `batch_wire`: atomic mode (default) validates all entries first, then applies all writes in a single Undo group
- Shared helpers: `TryGetComponentAndObjectReferenceProperty`, `ResolveReference`, `ValidateReferenceAssignment`

### Design Decisions

- **Extends existing tool** rather than new top-level tools — follows the `manage_*` multi-action pattern
- **SerializedProperty** for writes (not raw reflection) — correct dirty state, prefab overrides, Undo
[...]

PR #1362, opened 2026-09-01
ci: report Skipped instead of a green check when Unity tests cannot run

Scriptwonder:
Both Unity workflows have been reporting a **green check having run nothing** on every fork PR. Since practically every PR to this repo comes from a fork, that is most of the queue.

### The mechanism

Both gate every real step on a step-level `if:` reading an in-job license-detection output:

```yaml
- name: Run tests
  if: steps.detect.outputs.unity_ok == 'true'
```

A step-level `if:` yields step-conclusion `skipped`, which contributes **nothing** to the job conclusion. So the job exits clean and the merge box shows a tick.

Both workflows already knew. From their own source, before this PR:

```
# the detect step downstream writes unity_ok=false and the
# job reports a green check having compiled and tested nothing.
```
```
# Every step below is gated on unity_ok, so the job reports a green check
# having run nothing at all. Say so plainly on the run page.
```

The step summary does say so. But a step summary is not what a reviewer scans — the tick is.

### What it looked like in practice

PR #1323 does not compile. `SkyboxOps.cs(18,17): error CS0266`, red on all three OS since 2026-08-15. Here is how it presents:

```
Compile MCPForUnity (win/osx/linux)   = FAILURE   ← the one honest check
Test in editmode on Unity 6000.0.75f1 = SUCCESS   ← ran nothing
e2e-bridge                            = SUCCESS   ← ran nothing
Build site                            = SUCCESS
Compute Unity version matrix          = SUCCESS
CodeRabbit                            = SUCCESS
```

Five green ticks on a PR that provably does not build. The same shape appears on #1285, #1287 and #1043, where the two hollow ticks are 100% of their green.

### The fix

Hoist detection into a `license` gate job and gate the real job on it with a **job-level** `if:`:

```yaml
  license:
    outputs:
      unity_ok: ${{ steps.detect.outputs.unity_ok }}

  testAllModes:
    needs: [matrix, license]
    if: needs.license.outputs.unity_ok == 'true'
```

A job skipped by a job-level `if:` reports **Skipped** in the merge box, which is visually distinct from a pass. Branch protection [treats a skipped required check as satisfied](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches), so **this does not block merges** — it only stops the check claiming a pass it never earned.

No change to what runs when secrets *are* in scope: same steps, same `actions/cache` key, same matrix, same artifacts.

### Also

- Documents which checks are real signal on a fork PR, in `website/docs/contributing/dev-setup.md`:

  | Check | On a fork PR |
  |---|---|
  | `Compile MCPForUnity (win/osx/linux)` | **Runs** — real signal |
  | `Run Python Tests` | **Runs** — real signal |
  | `Check docs reference is fresh` | **Runs** — real signal |
  | `Test in editmode on Unity <version>` | **Skipped** |
  | `e2e-bridge` | **Skipped** |

- Surfaces the escape hatch that was previously buried in a workflow comment: to get a real Unity run against fork code, a maintainer pushes the reviewed branch into this repo and the `push` trigger runs the full suite in a trusted context. (There is deliberately no `pull_request_target` for these — fork-authored C# plus `UNITY_*` secrets is the classic "pwn request" shape.)
- Drops the stale `safe-to-test` references from the contributor docs; that gate was retired in #1308.

### Testing

Both files validate as YAML and the job graph is as intended (`matrix` → `license` → `testAllModes`; `license` → `e2e-bridge`). Touches no C# and no Python. The real proof is this PR's own checks: it comes from a fork, so `Test in editmode` and `e2e-bridge` should now read **Skipped** rather than green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Unity validation workflows now report **Skipped** when required license secrets are unavailable, instead of showing a misleading successful check.
  * Simplified workflow execution and artifact handling for licensed test runs.
  * Unity checks now run only for changes within the relevant Unity-scoped paths.

* **Documentation**
  * Updated contributor guidance to explain fork pull request behavior, secret availability, trusted runs, and path-filtered Unity validation.

### DIFF
--- a/.github/workflows/e2e-bridge.yml
+++ b/.github/workflows/e2e-bridge.yml
@@ -28,9 +28,18 @@ env:
   UNITY_IMAGE: unityci/editor:ubuntu-2021.3.45f2-linux-il2cpp-3
 
 jobs:
-  e2e-bridge:
-    runs-on: ubuntu-24.04
-    timeout-minutes: 40
+  license:
+    # Hoisted out of the smoke job so the gate is a JOB-level condition. A step-level `if:`
+    # produces step-conclusion `skipped`, which contributes nothing to the job conclusion,
+    # so this workflow reported a green check having booted no Editor and exercised no tool
+    # call. A job skipped by a job-level `if:` reports "Skipped" in the merge box instead.
+    # Branch protection treats a skipped required check as satisfied, so merges are not blocked.
+    name: Detect Unity license secrets
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    outputs:
+      unity_ok: ${{ steps.detect.outputs.unity_ok }}
     steps:
       - name: Detect Unity license secrets
         id: detect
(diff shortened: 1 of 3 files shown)


Scriptwonder:
Self-review — the fix is confirmed by this PR's own checks (it comes from a fork, so it exercises exactly the case in question):

```
e2e-bridge                                 = SKIPPED   ← was SUCCESS
Test in ${{ matrix.testMode }} on Unity …  = SKIPPED   ← was SUCCESS
Detect Unity license secrets               = SUCCESS   ← the new gate job
```

**One caveat worth checking before merge**, which I did not anticipate and only saw in the result above.

A matrix job skipped by a job-level `if:` reports under its **unexpanded** name — `Test in ${{ matrix.testMode }} on Unity ${{ matrix.unityVersion }}` — because the matrix values were never substituted. When it actually runs it still reports as `Test in editmode on Unity 6000.0.75f1`.

So the check name is no longer stable between the run and skip cases. If `Test in editmode on Unity 6000.0.75f1` is configured as a **required status check** on `beta`, a skipped run would surface as that required check being *absent* rather than skipped, which can leave a PR pending forever instead of mergeable. I don't have admin on the repo so I can't read the branch-protection config to check.

Two options:

1. **If it isn't a required check** — merge as-is. The unexpanded name is cosmetic.
2. **If it is** — give the job a static `name:` (e.g. `Unity EditMode tests`) so the check name is identical whether it runs or skips, with the version staying visible inside the run. That needs a one-time update to the required-checks list.

Happy to push option 2 if you tell me which it is.

PR #1370, opened 2026-09-02
fix(execute_code): allow using directives at the top of a snippet

lgarczyn:
## Description
A `using` directive at the top of an `execute_code` snippet fails to compile, because the snippet is spliced into a method body where the directive is illegal. This hoists a leading run of directives into the wrapper header instead.

## Type of Change
- [x] New feature (non-breaking change that adds functionality)
- [ ] Documentation update
- [x] Test update

## Changes Made
- `WrapUserCode` hoists the leading run of `using` directives into the wrapper header. Comments and blank lines are skipped; the first real statement stops the scan.
- Hoisted lines are blanked in place rather than removed, so compiler line numbers still map to what the caller wrote.
- The line offset is now derived from the generated header instead of a hardcoded `WrapperLineOffset = 10`, which silently drifted whenever the header changed. It is threaded through `CodeDomCompile` and `RoslynCompiler.Compile`.
- Directives, `using static`, and aliases (including generic and array targets) are recognised. The C# 8 `using var x = ...;` declaration deliberately is not, since hoisting it would break the body.
- 6 EditMode tests in `ExecuteCodeTests.cs`.

- Unity version(s) tested: 2022.3.62f2, LinuxEditor (headless)
- Package source used (`#beta`, `#main`, tag, branch, or `file:`): `file:` — branch checked out and symlinked in as an embedded package
- Resolved commit hash from `Packages/packages-lock.json` (if using a Git package URL): n/a, `"source": "embedded"`. Branch tested at `3722415d`.

- [ ] Unity EditMode tests
- [ ] Unity PlayMode tests

The 6 new EditMode tests are deliberately not ticked: they have not been run through the Test Runner. Every case they assert was exercised against a live editor instead, see Additional Notes. No Python changed.

## Documentation Updates

No tool or resource is added, removed, or re-signatured. One caveat below.

## Related Issues
None.

## Additional Notes

**How this was verified.** The 6 assertions were run as live `execute_code` calls against a headless 2022.3.62f2 editor with this branch loaded as the package, under both compiler backends:

| case | `auto` | forced `codedom` |
|---|---|---|
| leading `using System.IO;` hoisted | pass | pass |
| `using static System.Math;` | pass | pass |
| generic alias `using L = List<int>;` | pass | pass |
| `using` after a leading comment | pass | pass |
| redundant builtin `using System;` not duplicated | pass | pass |
| error line still points at caller line 3 | pass | pass |

Both backends matter because on this branch alone `auto` resolves to CodeDom, i.e. C# 6. An earlier draft included a `using var x = ...;` case, which is C# 8 and failed under CodeDom; it moved to the Roslyn PR below, where the language version is guaranteed.

**Doc caveat.** The `code` parameter is documented as "Must be a valid method body." That is now slightly narrow, since a leading run of `using` directives is accepted and hoisted. It is a one-line change in `execute_code.py` plus a docs regen: happy to add it here or leave it for a follow-up.

**Related PR.** `fix(execute_code): load Roslyn from Unity instead of falling back to C# 6` also touches `ExecuteCode.cs`. Whichever lands second needs a rebase. This one is the better base.

## Summary by CodeRabbit

- **New Features**
  - Execute Code now supports leading `using` directives, including `using static`, aliases, generic type aliases, and directives following comments.
  - Common `System.Text` functionality is available by default in executed snippets.

- **Bug Fixes**
  - Improved error reporting so line numbers correctly correspond to submitted code, even when using directives are present.
  - Prevented duplicate handling of standard built-in namespaces.
  - Strengthened safety checks to prevent namespace imports, aliases, or static imports from bypassing blocked-operation protections.

### DIFF
--- a/MCPForUnity/Editor/Tools/ExecuteCode.cs
+++ b/MCPForUnity/Editor/Tools/ExecuteCode.cs
@@ -529,20 +529,46 @@ namespace MCPForUnity.Editor.Tools
         private static bool IsUsingDirective(string line)
             => UsingDirectiveLine.IsMatch(line) || UsingAliasLine.IsMatch(line);
 
+        // A directive can carry comments, but the regexes above must see a bare `using ...;`
+        private static string StripComments(string line, ref bool inBlock)
+        {
+            var sb = new StringBuilder();
+            for (int i = 0; i < line.Length; i++)
+            {
+                bool pair = i + 1 < line.Length;
+                if (inBlock)
+                {
+                    if (pair && line[i] == '*' && line[i + 1] == '/') { inBlock = false; i++; }
+                    continue;
+                }
+                if (pair && line[i] == '/' && line[i + 1] == '*') { inBlock = true; i++; continue; }
+                if (pair && line[i] == '/' && line[i + 1] == '/') break;
+                sb.Append(line[i]);
+            }
+            return sb.ToString();
+        }
+
+        // Blanked, not removed, so compiler line numbers still map to the caller's
+        private static List<string> HoistLeadingUsings(string[] body)
+        {
+            var hoisted = new List<string>();
+            bool inBlock = false;
+            for (int i = 0; i < body.Length; i++)
+            {
+                string directive = StripComments(body[i], ref inBlock).Trim();
+                if (directive.Length == 0) continue;
+                if (!IsUsingDirective(directive)) break;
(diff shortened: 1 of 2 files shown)


Scriptwonder:
Hi, thanks for the PR. IF we have using System.IO; /* start, would the whole line including the /* gets deleted as well and cause a regression?

PR #1374, opened 2026-09-02
fix(execute_code): load Roslyn from Unity instead of falling back to C# 6

lgarczyn:
## Description

`execute_code` advertises a Roslyn backend, but on a stock editor
`Microsoft.CodeAnalysis` is not loaded, so `auto` silently falls through to the CodeDom
provider and the snippet is compiled as C# 6. Nothing reports this — the call just fails
on any newer syntax, and the caller has no way to tell why.

Unity already ships a Mono-loadable Roslyn at
`EditorApplication.applicationContentsPath/Tools/ScriptUpdater`, used by its own script
updater. Loading that gets a working Roslyn with nothing installed.

The `DotNetSdkRoslyn` copy sitting beside it is .NET Core and `BadImageFormat`s in the
editor domain, so it is deliberately not used.

## Type of Change

- [x] Documentation update
- [ ] Test update

## Changes Made

`MCPForUnity/Editor/Tools/ExecuteCode.cs`
- `TryLoadUnityRoslyn()`: when `Microsoft.CodeAnalysis.CSharp` is absent, `Assembly.LoadFrom`
  the two assemblies out of Unity's `Tools/ScriptUpdater`.

`Server/src/services/tools/execute_code.py`
- The description promised `C# 12+` and told callers `Microsoft.CodeAnalysis` had to be
  installed. Neither holds now: nothing needs installing, and the editor's copy is Roslyn
  3.11, which caps the language at C# 9. Both the tool description and the `compiler`
  parameter annotation were corrected.

`website/docs/reference/tools/scripting_ext/execute_code.md`
- Regenerated with `tools/generate_docs_reference.py`. It is the only page that changed;
  `--check` is clean afterwards.

- Unity version(s) tested: 2022.3.62f2
- Package source used: `file:` (local checkout)
- Resolved commit hash from `Packages/packages-lock.json`: n/a (local package)

The load is best-effort and wrapped in try/catch. If the path is missing on some editor
version, the behaviour is exactly what it is today: the CodeDom fallback.

- [x] Python tests (`cd Server && uv run pytest tests/ -v`) — 1374 passed, 3 skipped
- [x] Unity EditMode tests
- [ ] Unity PlayMode tests

`ExecuteCodeTests`: 46 total, 46 passed, on 2022.3.62f2. Two are new:

- `Execute_Auto_UsesRoslynLoadedFromUnity` — asserts `compiler == "roslyn"`, so a silent
  regression back to CodeDom fails the build instead of quietly capping the language.
- `Execute_UsingDeclaration_CompilesUnderRoslyn` — a C# 8 `using var`, unavailable on
  CodeDom, which pins the language version rather than just the backend name.

## Documentation Updates

- [x] I have added/removed/modified tools or resources
- [x] If yes, I have updated all documentation files using:
  - [x] The LLM prompt at `tools/UPDATE_DOCS_PROMPT.md` (recommended)
  - [x] Manual review of the generated changes

No tool was added or removed, so `manifest.json` and the README tool lists are unchanged.

## Related Issues

None.

## Additional Notes

`tools/UPDATE_DOCS_PROMPT.md` step 4 tells you to run `python3 tools/check_docs_sync.py`.
That script does not exist in the repo. `tools/generate_docs_reference.py --check` appears
to be the current gate — worth fixing that line separately.

## Summary by CodeRabbit

- **New Features**
  - The `execute_code` tool can now use the Roslyn compiler bundled with the Unity Editor when using `auto` or `roslyn`.
  - C# language support is available through C# 9, including modern syntax such as using declarations.
  - The `codedom` option remains available for legacy C# 6 compilation.

- **Documentation**
  - Updated tool descriptions and compiler-option guidance to accurately reflect compiler behavior and supported language versions.

- **Tests**
  - Added coverage confirming automatic Roslyn selection and successful compilation of C# 8 using declarations.

### DIFF
--- a/MCPForUnity/Editor/Tools/ExecuteCode.cs
+++ b/MCPForUnity/Editor/Tools/ExecuteCode.cs
@@ -663,10 +663,32 @@ namespace MCPForUnity.Editor.Tools
             _isAvailable = null;
         }
 
+        // Unity's script updater ships a Mono-loadable Roslyn, so roslyn needs nothing installed
+        // The DotNetSdkRoslyn copy beside it is .NET Core and BadImageFormats in the editor domain
+        private static void TryLoadUnityRoslyn()
+        {
+            try
+            {
+                string dir = Path.Combine(UnityEditor.EditorApplication.applicationContentsPath, "Tools", "ScriptUpdater");
+                foreach (string name in new[] { "Microsoft.CodeAnalysis.dll", "Microsoft.CodeAnalysis.CSharp.dll" })
+                {
+                    string path = Path.Combine(dir, name);
+                    if (File.Exists(path)) Assembly.LoadFrom(path);
+                }
+            }
+            catch (Exception e)
+            {
+                McpLog.Warn($"[ExecuteCode] Could not load Unity's Roslyn: {e.Message}");
+            }
+        }
+
         private static bool Initialize()
         {
             try
             {
+                if (Type.GetType("Microsoft.CodeAnalysis.CSharp.CSharpSyntaxTree, Microsoft.CodeAnalysis.CSharp") == null)
+                    TryLoadUnityRoslyn();
+
                 _syntaxTreeType = Type.GetType("Microsoft.CodeAnalysis.CSharp.CSharpSyntaxTree, Microsoft.CodeAnalysis.CSharp");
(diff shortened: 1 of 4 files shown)


Scriptwonder:
Hi thanks for the PR, can you run double check on how the hardcoded ScriptUpdater path changes across versions? I think UNity 6 Roslyn is moved to a different location that your hardcoded path, so the same problem will not be resolved.

